PR #18, opened 2020-05-18
Rename package

charleskawczynski:
It's been discussed that this package is useful beyond the moist atmosphere, and we should generalize the name to reflect this. This PR changes the package name from `MoistThermodynamics.jl` to `Thermodynamics.jl`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
-name = "MoistThermodynamics"
-uuid = "0460cc1c-04d8-4f02-b57d-511136c80085"
+name = "Thermodynamics"
+uuid = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 authors = ["Charles Kawczynski <kawczynski.charles@gmail.com>"]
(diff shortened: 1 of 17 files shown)


stillyslalom:
MoistThermodynamics.jl may have been too narrow, but Thermodynamics.jl seems too broad - the intent seems to be a tight coupling with CliMA, and there's a lot of thermodynamics happening outside of [-15-45 °C water/air mixtures]. Compare to [thermo](https://thermo.readthedocs.io/en/latest/) or [CoolProp](http://www.coolprop.org/) for breadth. How about AtmosphericThermo.jl, GeophysicalThermo.jl, or ClimateThermo.jl?

charleskawczynski:
Hi @stillyslalom. Thanks for your suggestion. At the moment, Thermodynamics.jl should probably be renamed to AtmosphericThermo.jl, but there has been discussion to include thermo support for ocean and land, so we may put off renaming for now but we should revisit this.

PR #36, opened 2021-06-04
Make tested profiles accessible in submodule

charleskawczynski:
ClimateMachine.jl [uses Thermodynamics.jl's tested profiles](https://github.com/CliMA/ClimateMachine.jl/blob/master/test/Atmos/prog_prim_conversion/runtests.jl#L110) for primitive-prognostic conversions. This PR makes those profiles available via a submodule in `src/`.


charleskawczynski:
This PR, or an alternative solution, is needed for ClimateMachine.jl to upgrade to use Thermodynamics.jl (which is in progress in [this PR](https://github.com/CliMA/ClimateMachine.jl/pull/2209)).

PR #38, opened 2021-07-27
Add psq constructor

charleskawczynski:
This PR adds a new constructor, `PhaseEquil_psq`, and tests it for the newly added `entropy` profiles.

In addition, this PR add `air_density_equil`, which seems to be important for the consistency of `PhaseEquil_psq` and may be useful elsewhere.

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -58,6 +58,7 @@ export specific_enthalpy
 export total_specific_enthalpy
 export moist_static_energy
 export saturated
+export entropy
 
 heavisided(x) = (x > 0) * x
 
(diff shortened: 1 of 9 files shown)


src/Thermodynamics.jl
```diff
@@ -50,6 +50,9 @@ using KernelAbstractions: @print
 using CLIMAParameters: AbstractParameterSet
 using CLIMAParameters.Planet
 const APS = AbstractParameterSet
+entropy_dry_air(::APS) = 6864.8 # J / (kg K)
+entropy_water_vapor(::APS) = 10513.6 # J / (kg K)
+standard_temperature(::APS) = 298.15 # K
```
charleskawczynski, line 55: These need to be moved to CLIMAParameters.jl, waiting on review / merging of [CLIMAParameter.jl's #51](https://github.com/CliMA/CLIMAParameters.jl/pull/51).

src/relations.jl
```diff
@@ -145,6 +146,94 @@ function air_density(
     return p / (gas_constant_air(param_set, q) * T)
 end
 
+"""
+    air_density_equil(
```
tapios, line 150: Why do we need this? That is, why can't we just get the density from temperature, specific humidities, and whatever other thermodynamic variable (energy, entropy etc.) we are using?

tapios (commented):
Looks ok to me. But why are we building all this machinery for entropy, which we will not need in the end? I'm concerned it creates code bloat without a clear purpose in the end. (If it's just for testing the parameterization for now, we can leave it in the parameterization rather than adding to Thermodynamics.)

src/relations.jl
charleskawczynski: It seems that we need this in `saturation_adjustment_psq` because temperature is not yet known. I tried using `air_density(..., PhasePartition(q_tot))`, but this leads to consistency failures in the following tests:

```julia
ts_psq = PhaseEquil_psq.(param_set, p, s, q_tot)
@test all(entropy.(ts_psq) .≈ s)                                     # requires air_density_equil
@test all(isapprox.(air_pressure.(ts_psq), p; rtol = rtol_pressure)) # requires air_density_equil
```

Since this is only used for the reference state in TC.jl, I don't think performance is much of an issue. I think this function (`air_density_equil`) could fix consistency errors in other thermo functions, but we're already accurate to the tolerances in the test suite, so I don't think it's much of an issue.

The reason I went down this rabbit hole was because, IIRC, without `air_density_equil`, the TC.jl reference states are slightly different and one or two of the simulations crashed.

charleskawczynski:
> Looks ok to me. But why are we building all this machinery for entropy, which we will not need in the end? I'm concerned it creates code bloat without a clear purpose in the end. (If it's just for testing the parameterization for now, we can leave it in the parameterization rather than adding to Thermodynamics.)

Fair point. We can certainly leave this in TC.jl for now if you'd prefer.

src/relations.jl
tapios: We can reformulate this without another iterative method. But better yet, we can do without any of it. No need to entropy-based reference states.

tapios:
> > Looks ok to me. But why are we building all this machinery for entropy, which we will not need in the end? I'm concerned it creates code bloat without a clear purpose in the end. (If it's just for testing the parameterization for now, we can leave it in the parameterization rather than adding to Thermodynamics.)
> 
> Fair point. We can certainly leave this in TC.jl for now if you'd prefer. I don’t have a strong preference on where the code lives :)

Let's remove the entropy-based reference states. They are a holdover from PyCLES's anelastic approximations, but they are not necessary for the single-column model. Let's modify the test cases that use these reference states to simpler reference states.

charleskawczynski:
> Let's remove the entropy-based reference states. They are a holdover from PyCLES's anelastic approximations, but they are not necessary for the single-column model. Let's modify the test cases that use these reference states to simpler reference states.

I had tried using the existing thermo methods ([here](https://github.com/CliMA/TurbulenceConvection.jl/pull/25)), but several cases had issues (and Soares was crashing) as a result of the different reference state. You can see the results by clicking into the buildkite link, then clicking on Artifacts for each case. The constant entropy warnings were printed in a lot of those cases, and so @yairchn and I thought that a constant-entropy constructor might be necessary.

The attempt to use a [constant entropy reference state](https://github.com/CliMA/TurbulenceConvection.jl/pull/43) is yielding a stable Soares, albeit TRMM still has some undesirable changes.

I'm open to any ideas on how to move forward.

tapios:
> > Let's remove the entropy-based reference states. They are a holdover from PyCLES's anelastic approximations, but they are not necessary for the single-column model. Let's modify the test cases that use these reference states to simpler reference states.
> 
> I had tried using the existing thermo methods ([here](https://github.com/CliMA/TurbulenceConvection.jl/pull/25)), but several cases had issues (and Soares was crashing) as a result of the different reference state. You can see the results by clicking into the buildkite link, then clicking on Artifacts for each case. The constant entropy warnings were printed in a lot of those cases, and so @yairchn and I thought that a constant-entropy constructor might be necessary.
> 
> The attempt to use a [constant entropy reference state](https://github.com/CliMA/TurbulenceConvection.jl/pull/43) is yielding a stable Soares, albeit TRMM still has some undesirable changes.
> 
> I'm open to any ideas on how to move forward.

Pease show the results tomorrow and let's talk about it then. The reference state should not be crucial, and if it is, it points to other problems.

charleskawczynski:
Superseded by #52.

PR #49, opened 2021-09-14
Adding pressure as a variable in Thermodynamic state

yairchn:
To accommodate the requirements for Anelastic, or even for SGS Anelastic, assumptions of the EDMF, we need to allow the Thermodynamic state to store the pressure variable given as input rather than computing it from the state variables when need. This will allow us to apps in either the reference (hydrostatic) pressure or the grid mean pressure make sure that this pressure is used consistently in that EDMF subdomain.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["Climate Modeling Alliance"]
 version = "0.4.0"
 
 [deps]
+ArtifactWrappers = "a14bc488-3040-4b00-9dc1-f6467924858a"
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 ExprTools = "e2ba6199-217a-4e67-a87a-7c52f15ade04"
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Climate Modeling Alliance"]
 version = "0.4.0"
 
 [deps]
+ArtifactWrappers = "a14bc488-3040-4b00-9dc1-f6467924858a"
```
charleskawczynski, line 7: This file should not have been edited for this PR, and we shouldn't be adding ArtifactWrappers to the Project.toml

src/states.jl
```diff
@@ -96,9 +96,15 @@ struct PhaseDry{FT, PS} <: ThermodynamicState{FT}
     e_int::FT
     "density of dry air"
     ρ::FT
+    "air pressure"
+    p::FT
```
charleskawczynski, line 100: Is this needed? I don't think it is.

yairchn: but the call or 'air_pressure' in relations/jl line 119:
air_pressure(ts::ThermodynamicState) = ts.p
is for all Phases. Without this it will break in tests

PR #50, opened 2021-09-14
Make thermodynamic state more extendable

charleskawczynski:
We'll now be able to make our own thermo state in TC.jl that includes pressure, so that we don't bloat the data structures when not needed. Supersedes #49.


yairchn (approved):
LGTM

PR #52, opened 2021-09-15
Add pressure to ts and make PhaseEquil_pθq more consistent

yairchn:
This is the final version of our attempts to add an Anelastic constructor with minimal changes in the Repo.
We are adding a new Phase_type 'AnelasticPhaseEquil'
A new constructor 'AnelasticPhaseEquil_pθq'
A new saturation adjustment 'saturation_adjustment_given_pθq_anelastic'
A new method 'PhasePartition_equil_given_p'
And liquid_fraction now dispatches the to see if liquid fraction is 1 (liquid only) for the 'AnelasticPhaseEquil'
We dispatch 'air_pressure' to return the input (hydrostatic or grid mean) pressure.


charleskawczynski (commented):
Need to update the docs. See the doc build error messages, [this](https://github.com/CliMA/ClimateMachine.jl/blob/master/.github/workflows/doc_build_common_error_messages.md) is a doc error message cheat sheet I've found useful in the past.

docs/src/API.md
```diff
@@ -12,6 +12,7 @@ Thermodynamics
 ```@docs
 PhasePartition
 PhasePartition_equil
+PhasePartition_equil_given_p
```
charleskawczynski, line 15: We can remove this. This list is only for external methods.

src/relations.jl
```diff
@@ -1015,6 +1033,27 @@ function q_vap_saturation_from_pressure(
     return p_v_sat / (ρ * _R_v * T)
 end
 
+"""
+    q_vap_saturation_from_partial_pressures(param_set, q_tot, p, p_v_sat)
```
charleskawczynski, line 1037: Let's update this doc string to match the code.

src/relations.jl
```diff
@@ -1015,6 +1033,27 @@ function q_vap_saturation_from_pressure(
...
+Compute the saturation specific humidity, given
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `q_tot` total water specific humidity,
+ - `p` air pressure,
+ - `p_v_sat` saturation vapor pressure
```
charleskawczynski, line 1042: Let's update this list to match the code args

src/relations.jl
```diff
@@ -1800,6 +1908,117 @@ function saturation_adjustment_given_pθq(
...
+ - `q_tot` total specific humidity
+ - `phase_type` a thermodynamic state type
+ - `tol` absolute tolerance for saturation adjustment iterations. Can be one of:
+    - `SolutionTolerance()` to stop when `|x_2 - x_1| < tol`
+    - `ResidualTolerance()` to stop when `|f(x)| < tol`
+ - `maxiter` maximum iterations for non-linear equation solve
+
+by finding the root of
+
+`θ_{liq_ice} - liquid_ice_pottemp_sat(param_set, T, ρ, phase_type, q_tot) = 0`
```
charleskawczynski, line 1936:
```suggestion
`θ_{liq_ice} - liquid_ice_pottemp_given_pressure(param_set, T, ρ, phase_type, q_tot) = 0`
```

We should probably add a note that this saturation adjustment assumes anelastic.

src/states.jl
```diff
@@ -491,6 +493,88 @@ function PhaseEquil_ρpq(
...
+
+# Constructors
+
+    PhaseEquil(param_set, ρ, p, e_int, q_tot, T)
+
+# Fields
+
+$(DocStringExtensions.FIELDS)
+"""
+
```
charleskawczynski, line 514:
```suggestion
```
Otherwise this doc string is broken

charleskawczynski (commented):
Overall looks good! I added comments for some things to be fixed / cleaned up.

Since this will be used heavily in TurbulenceConvection.jl, we should probably have a few eyes scan these equations for correctness.

src/relations.jl
trontrytel: Should we mention something about it being added for the purpose of being compatible with anelastic approximation here?

src/relations.jl
trontrytel: yes, please!

src/states.jl
```diff
@@ -491,6 +493,88 @@ function PhaseEquil_ρpq(
...
+struct AnelasticPhaseEquil{FT, PS} <: AbstractPhaseEquil{FT}
+    "parameter set, used to dispatch planet parameter function calls"
+    param_set::PS
+    "density of air (potentially moist)"
```
trontrytel, line 518: maybe just to be extra clear we could write here that it's not the reference density?

trontrytel (approved):
This looks good to me. With the added functionality I should be able to couple CloudMicrophysics.jl to TC.jl without too many changes in the microphysics package.

tapios (commented):
My general comment is to avoid reference to anelastic if possible. Most changes here just use pressure as a thermodynamic variable, without that having to be an anelastic reference pressure.

src/relations.jl
```diff
@@ -1015,6 +1033,30 @@ function q_vap_saturation_from_pressure(
...
+Alternative method to compute the saturation specific humidity without using density, given
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `q_tot` total water specific humidity,
+ - `p` air pressure,
+ - `T` air tempearture
+-  `phase` Phase
+
+This is useful for anelastic model where air_density should be avioded
+"""
+function q_vap_saturation_from_partial_pressures(
```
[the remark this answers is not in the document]
yairchn: I agree, however the reason I choose this name is that we already have a function name 'q_vap_saturation_from_pressure' which to my opinion is named badly as it uses only p_v_sat:

https://github.com/CliMA/Thermodynamics.jl/blob/304e5c12bdfb80f1a1c9a2b62d6d6f01b2c9315e/src/relations.jl#L1008

I guess we can rename both to make their names more consistent.

src/relations.jl
```diff
@@ -1800,6 +1911,119 @@ function saturation_adjustment_given_pθq(
     end
 end
 
+"""
+    saturation_adjustment_given_pθq_anelastic(
```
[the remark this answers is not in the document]
yairchn: We already have a function name 'saturation_adjustment_given_pθq'
https://github.com/CliMA/Thermodynamics.jl/blob/304e5c12bdfb80f1a1c9a2b62d6d6f01b2c9315e/src/relations.jl#L1735

what makes this new one Anelastic is the fact that it avoids calling 'air_density', it such a call will return the buoyancy density not the reference one. 

I am happy to accommodate any other suggested name of course.

src/states.jl
```diff
@@ -491,6 +493,87 @@ function PhaseEquil_ρpq(
...
+"""
+struct AnelasticPhaseEquil{FT, PS} <: AbstractPhaseEquil{FT}
+    "parameter set, used to dispatch planet parameter function calls"
+    param_set::PS
+    "full density of air (potentially moist)"
+    ρ::FT
+    "reference air pressure"
+    p::FT
+    "internal energy"
+    e_int::FT
```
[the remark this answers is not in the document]
yairchn: Currently we do not need it but if we change the prognostic variable in the EDMF to energy, this will be needed. therefore I left it in.
If you still think it should be removed from the thermodynamic state we can take it.

yairchn (commented):
Thanks @tapios , see replies to your comments

src/relations.jl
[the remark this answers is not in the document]
tapios: I suggest you rename `q_vap_saturation_from_pressure` to `q_vap_saturation_given_density` and then call your new function `q_vap_saturation_given_pressure`.

src/relations.jl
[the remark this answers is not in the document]
tapios: I'd just change `saturation_adjustment_given_pθq` to do what you need--I don't see why we need both functions. It's not a good idea anyway for `saturation_adjustment_given_pθq` to call `air_density`.

src/states.jl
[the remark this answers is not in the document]
tapios: Again, I'd like to get rid of references to anelastic in thermodynamics (because it really doesn't belong there). If we use internal or total energy, and if it turns out we need a new constructor for thermodynamics, we can add it then. But here, all I think we need is a constructor for thermodynamics given pressure, and that does not require a new `PhaseEquil`, just modifications to functions that should take pressure as input (and not compute density along the way).

tapios (commented):
Again, I'd like to get rid of references to anelastic in thermodynamics (because it really doesn't belong there). All we need should be thermodynamic calculations with pressure as a state variable, without computing density along the way. I think we should be able to achieve that without new constructors etc., just by modifying the existing calculations with given pressure.

src/relations.jl
charleskawczynski: Let's add temperature to the argument list

charleskawczynski: And replace `p_v_sat` with `phase`

src/relations.jl
```diff
@@ -1212,6 +1239,44 @@ PhasePartition_equil(ts::AbstractPhaseNonEquil) = PhasePartition_equil(
...
+    param_set::APS,
+    T::FT,
+    p::FT,
+    q_tot::FT,
+    phase_type::Type{<:ThermodynamicState},
+) where {FT <: Real}
+
+    q_v_sat = q_vap_saturation_from_pressure(param_set, q_tot, p, T, Liquid())
+    _liquid_frac = liquid_fraction(param_set, T, phase_type)                    # fraction of condensate that is liquid
+    q_c = q_tot - q_v_sat
```
ilopezgp, line 1264: Since you are computing q_c from this q_v_sat, I believe that the p_vap_saturation has to be computed using the weighted average of latent heats, with the constructor 

`saturation_vapor_pressure(
    param_set::APS,
    phase_type::Type{<:ThermodynamicState},
    T::FT,
)`

Right now you are computing the q_vap_sat for liquid.

src/relations.jl
```diff
@@ -1651,16 +1716,16 @@ end
...
     )
 
 Compute the temperature `T` that is consistent with
 
  - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `p` air pressure
  - `θ_liq_ice` liquid-ice potential temperature
  - `q_tot` total specific humidity
- - `p` pressure
  - `phase_type` a thermodynamic state type
```
ilopezgp, line 1729: Switch `max_iter` and `tol`in docstring.

src/states.jl
```diff
@@ -217,6 +217,8 @@ struct PhaseEquil{FT, PS} <: AbstractPhaseEquil{FT}
     param_set::PS
     "density of air (potentially moist)"
     ρ::FT
+    "reference air pressure"
+    p::FT
```
ilopezgp, line 221: By reference what do you mean, exactly?

src/states.jl
```diff
@@ -262,7 +264,16 @@ function PhaseEquil_ρeq(
         maxiter,
         temperature_tol,
     )
-    return PhaseEquil{FT, typeof(param_set)}(param_set, ρ, e_int, q_tot_safe, T)
+    q_pt = PhasePartition_equil(param_set, T, ρ, q_tot, phase_type)
```
ilopezgp, line 267: `q_tot` or `q_tot_safe` ?

src/states.jl
yairchn: its probably a rebasing error we decided to remove any indication of anelastic or reference

src/relations.jl
yairchn: great catch. I guess that what you mean is that 'saturation_vapor_pressure' used inside 'q_vap_saturation_from_pressure' should be changed here? 
https://github.com/CliMA/Thermodynamics.jl/blob/7a560298c1f49288ad30b74b54c2a8c5392e0f8d/src/relations.jl#L1040

yairchn:
@ilopezgp thanks for the great review. I am updating the branch. Regarding the comment on 'saturation_vapor_pressure' I am pushing the change I think is needed in line 1040 in src/relations.jl , with the original call there in comment. let me know if we agree on that change and will clean the comment

src/relations.jl
```diff
@@ -1212,6 +1237,42 @@ PhasePartition_equil(ts::AbstractPhaseNonEquil) = PhasePartition_equil(
...
+
+    q_v_sat = q_vap_saturation_from_pressure(param_set, q_tot, p, T, phase_type)
+    _liquid_frac = liquid_fraction(param_set, T, phase_type)
+    q_c = q_tot - q_v_sat
+    q_liq = _liquid_frac * q_c
+    q_ice = (1 - _liquid_frac) * q_c
+    return PhasePartition(q_tot, q_liq, q_ice)
+end
+
+PhasePartition_given_p(ts::PhaseEquil) = PhasePartition_equil_given_p(
```
charleskawczynski, line 1268: This is the one big part of this PR that I'm uncomfortable with. This function should really be called `PhasePartition`-- but then that leaves us with a very opaque view of what's happening. Sometimes `PhasePartition_equil` is called, and other times `PhasePartition_equil_given_p` is called, and it's not clear from which constructors which route is taken. 

The way that we've got around this so far is by having one method call the other, and the method that implements the actual relation is the "core" method. This is really nice because it ensures that all relations we use are synchronized / consistent.

And the core problem with making `PhasePartition_equil_given_p` the core method is that, when pressure is _not_ given, we can't construct it when determining the phase partition (because it _depends_ on the phase partition).

I've discussed with @yairchn that, perhaps we can put a bit more effort into ensuring that we cannot make the existing methods work.

charleskawczynski (approved):

src/relations.jl
charleskawczynski: Ok, I think we've resolved things. See [this comment](https://github.com/CliMA/Thermodynamics.jl/pull/52#issuecomment-930411505).

test/relations.jl
```diff
@@ -925,7 +971,7 @@ end
             rtol = rtol_temperature,
         ))
         @test all(compare_moisture.(ts, q_pt))
-        @test all(isapprox.(air_pressure.(ts), p, atol = atol_pressure))
+        @test all(air_pressure.(ts) .≈ p)
```
yairchn, line 974: now its exact!

yairchn:
@trontrytel - one of the major concerns we had with the changes is to make sure that when Microphysics (or any other package) calls PhasePartition(ts) from EDMF it would be consistent given the assumption that EDMF does for pressure. In the current code, since the Pressure is stored in the ThermodynamicState and the density was computed using this pressure calling  PhasePartition(ts) which effectively uses density remains consistent with the pressure used in the computation of the ThermodynamicState.

tapios (approved):
LGTM. Thanks for getting this to the end, @yairchn and @charleskawczynski!

Just one question: Can't we remove the file with the helper functions? I can see you wanted this for testing the changes. But given that they seem correct now, why do we need it? I just seems like clutter whose purpose will be hard to parse for future readers of the code.

charleskawczynski:
> Just one question: Can't we remove the file with the helper functions? I can see you wanted this for testing the changes. But given that they seem correct now, why do we need it? I just seems like clutter whose purpose will be hard to parse for future readers of the code.

I guess I wanted to keep it for piece of mind--it's nice to know that the two formulations are linked in case there are future changes. That said, I'm fine to remove it if that's preferred. Up to you, @tapios:
 - [ ] keep the additional test
 - [ ] remove the additional test

tapios:
> > Just one question: Can't we remove the file with the helper functions? I can see you wanted this for testing the changes. But given that they seem correct now, why do we need it? I just seems like clutter whose purpose will be hard to parse for future readers of the code.
> 
> I guess I wanted to keep it for piece of mind--it's nice to know that the two formulations are linked in case there are future changes. That said, I'm fine to remove it if that's preferred. Up to you, @tapios:
> 
> * [ ]  keep the additional test
> * [ ]  remove the additional test

It makes sense you wanted to test it. But now that that's done and passed, I'd favor removal: less code, less clutter--unless others want to keep it.

charleskawczynski:
No problem, removed.

PR #58, opened 2021-10-16
Expose numerical method type to PhaseEquil_pθq constructor

charleskawczynski:
@trontrytel is having convergence issues in the `PhaseEquil_pθq` when adding ice in TRMM. This PR exposes the numerical method so that she can try out RegulaFalsi method, which [has better convergence properties](https://clima.github.io/ClimateMachine.jl/previews/PR1885/DevDocs/Common/Thermodynamics/Thermodynamics/).

cc @tapios


src/relations.jl
```diff
@@ -1766,22 +1767,33 @@ function saturation_adjustment_given_pθq(
...
-        KA.@print(", θ_liq_ice=", θ_liq_ice)
-        KA.@print(", q_tot=", q_tot)
-        KA.@print(", T=", sol.root)
-        KA.@print(", maxiter=", maxiter)
-        KA.@print(", tol=", tol.tol, "\n")
-        error("Failed to converge with printed set of inputs.")
+        if print_warning()
+            KA.@print("-----------------------------------------\n")
+            KA.@print("maxiter reached in saturation_adjustment_given_pθq:\n")
+            KA.@print("    Method=SecantMethod")
```
trontrytel, line 1786: @charleskawczynski should we update the printed method name?

trontrytel (approved):

PR #59, opened 2021-10-20
Improve convergence plots

charleskawczynski:
This PR was pulled from [ClimateMachine.jl's 1885](https://github.com/CliMA/ClimateMachine.jl/pull/1885), which improves the convergence plots for values outside of the tested profiles.

@trontrytel, we could add these same plots for the other thermo constructors to get an idea of the convergence manifold.


charleskawczynski:
Alright, I think this is ready, I'll merge the less controversial parts now, and open a separate PR for getting RegulaFalsi working.

charleskawczynski:
Alright, I think this is ready, I'll merge the less controversial parts now, and open a separate PR for getting RegulaFalsi working.

PR #62, opened 2021-11-04
use const latent heat in specific entropy

trontrytel:
As discussed today, at least for now we should stick to latent heat constant with temperature


ilopezgp (approved):
LGTM

charleskawczynski (approved):

PR #63, opened 2021-11-04
bump version

trontrytel:


charleskawczynski (approved):

charleskawczynski:
Hmm, actually, should we update this as breaking changes?

trontrytel:
I don't think so. The change between Lv0 and Lv(T)  should not be that big for our purposes

charleskawczynski:
Fine by me, just note that it might break TC.jl's CI since it's so sensitive, but we can just open a PR that updates the MSE tables

PR #74, opened 2021-12-17
Avoiding dev methods in allocation analysis

charleskawczynski:
Our "dev only" methods (which accept keywords) seem to incur allocations, which is misleading for our allocations monitoring. This PR changes our allocation analysis to call the state constructors directly. It turns out that our `ρeq` constructor is decent, but still allocates about 125 KB.


charleskawczynski (commented):
bors r+

PR #76, opened 2021-12-20
Fix gpu deps with specific gpu env + compats

### LINKED ISSUE #75: Add gpu test environment
We should probably add a gpu test environment so that we can limit `compat`s for GPU-related packages to ensure CI doesn't randomly break from updates (which currently update with breaking and non-breaking version changes).

charleskawczynski:
This PR fixes CI and closes #75.


charleskawczynski:
Hmm, not sure why we're updating to CUDA 3.6.4 even though our gpuenv/ specifies a compat for CUDA 3.5.

charleskawczynski:
cc @simonbyrne, any idea how this might happen?

charleskawczynski:
Ah, maybe the depot. I'll try clearing it.

Issue #87, opened 2022-03-03
Add a more flexible phase partition function


trontrytel:
Our current phase partitioning in the equilibrium formulation assumes everything below freezing temperature is ice (see `liquid_fraction` in Thermodynamics.jl):

https://github.com/CliMA/Thermodynamics.jl/blob/5945e684e36d7759088097d00fe4562ceb9db0cf/src/relations.jl#L1168-L1176

To make it more flexible I would like to change it to:

https://github.com/CliMA/pycles/blob/eb231a5f57bd5ebfd7c4bc1f65ab5fda0295768a/Microphysics.pxd#L106-L119

Setting `Twarm=Tcold=T_freeze` would result in the current behaviour. But at the same time, it would give us more flexibility to test the impact of different (even if prescribed) phase partitioning.

charleskawczynski:
@trontrytel , can we close this?

trontrytel:
Yes! The PR was merged in, I should have closed it ages ago

PR #88, opened 2022-03-03
Add more flexible phase partition in equilibrium

trontrytel:
As mentioned in #87 I would like to have a phase partition function similar to what is used in PyCLES. 

I know that in the end we will be using a non-equilibrium formulation. But I think in the mean time, it would be good to assume the same phase partition in LES and single column models.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Thermodynamics"
 uuid = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 authors = ["Climate Modeling Alliance"]
-version = "0.5.11"
+version = "0.6.0"
 
 [deps]
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
@@ -11,7 +11,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RootSolvers = "7181ea78-2dcb-4de3-ab41-2b8ab5a31e74"
 
 [compat]
-CLIMAParameters = "0.1, 0.2, 0.3, 0.4"
+CLIMAParameters = "0.4"
 DocStringExtensions = "0.8.1"
 KernelAbstractions = "0.7.2"
 RootSolvers = "0.2, 0.3"
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -1163,7 +1163,7 @@ them.
 
 # ThermodynamicState
 Otherwise, phase equilibrium is assumed so that the fraction of liquid
-is a function that is 1 above `T_freeze` and goes to zero below `T_freeze`.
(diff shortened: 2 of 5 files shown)


charleskawczynski (approved):
Just bump the heuristic number (1395) in `test/relations.jl`

src/relations.jl
```diff
@@ -1172,7 +1172,15 @@ function liquid_fraction(
     q::PhasePartition{FT} = q_pt_0(FT),
 ) where {FT <: Real, phase_type <: ThermodynamicState}
     _T_freeze::FT = CPP.T_freeze(param_set)
-    return FT(T > _T_freeze)
+    _T_icenuc::FT = CPP.T_icenuc(param_set)
+
+    if T > _T_freeze
+        return FT(1)
+    elseif (T > _T_icenuc && T <= _T_freeze)
+        return (T - _T_icenuc) / (_T_freeze - _T_icenuc)
```
tapios, line 1180: Would you want to allow arbitrary powers of this fraction?

test/relations.jl
```diff
@@ -703,7 +703,7 @@ end
 
         @test count(air_temperature.(ts_lower) .== Ref(_T_freeze)) ≥ 217
         @test count(air_temperature.(ts_upper) .== Ref(_T_freeze)) ≥ 217
-        @test count(air_temperature.(ts_mid) .== Ref(_T_freeze)) ≥ 1395
+        @test count(air_temperature.(ts_mid) .== Ref(_T_freeze)) ≥ 1392
```
tapios, line 706: Strange to have these arbitrary numbers here. Fractions of total would be more justifiable.

tapios:
This looks basically fine. The smooth liquid fraction function in principle should also affect the gradient that appears in Newton solvers. It may be worth finally updating those functions.

tapios (approved):

test/relations.jl
charleskawczynski: Yeah, we're using numbers just to be precise, this is just a heuristic, and it tells us that we're in fact testing the branch where we return `T_freeze`.

charleskawczynski:
> The smooth liquid fraction function in principle should also affect the gradient that appears in Newton solvers for saturation adjustment

I agree with this!

trontrytel:
bors try

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Thermodynamics"
 uuid = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 authors = ["Climate Modeling Alliance"]
-version = "0.5.11"
+version = "0.6.0"
```
charleskawczynski, line 4: I suppose the compat entry below should be `CLIMAParameters = "0.4"`

trontrytel: done

PR #83, opened 2022-02-03
[WIP] piloting parameter refactor

odunbar:
## Purpose
To prototype the new parameter-from-file interface with Thermodynamics from discussions in #82 

!! Now depends on `https://github.com/CliMA/CLIMAParameters.jl#orad/refactor-parameters-v2` !!
- `add https://github.com/CliMA/CLIMAParameters.jl#orad/refactor-parameters-v2` in `test/Project.toml` (see PR [62](https://github.com/CliMA/CLIMAParameters.jl#orad/refactor-parameters-v2) )
## In this PR 
- Use the default parameter toml file in `orad/refactor-parameters-v2` . Prototype parameters are:
```TOML
 [descriptive_name]
 alias = "deprecated_short_name"
 value = 0.435
 type = "float"
 description = "where necessary this is added"
 ```
- use file parser utilities from `CLIMAParameters`, store parameters in a `Dict` as `alias => value`. 
- create a local parameter struct, pass this to other relevant modules/tests in place of CLIMAParameters
- call parameters in file by `param_set.alias` replacing `CPP.alias(param_set)`
### ALSO: 
- added TOML to `Project.toml` in `test/` 
- added compat of `0.4` to CLIMAParameters in gpuenv
- added `PackageSpec(name="CLIMAParameters",rev="orad/refactor-parameters-v2")` to `.buildkite/pipeline.yml`
## To do  
- currently all naming is done by `alias`  (i.e `param_set.alias`) until the descriptive names are set
- enforcing concrete Float64 type for tests
- All tests pass, except docs (will update last)

PR #92, opened 2022-04-20
Remove param_set cache from thermo states

charleskawczynski:
This PR removes the `param_set` cache from thermodynamic states. This is a pretty big API / breaking change, however, it will allow us to capture thermodynamic states, which seems like it will be both necessary and handy (we'll be able to make Thermo state fields, and hopefully remove several aux fields as a result).

I think that this PR and #91 will help shrink #83 into a more isolated PR.

@trontrytel, any idea on the blast radius this will have in Microphysics?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Thermodynamics"
 uuid = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 authors = ["Climate Modeling Alliance"]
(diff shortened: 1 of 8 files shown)


trontrytel (approved):
Looks good to me. I can take a pass on updating CloudMicrophysics.jl to those changes tomorrow. I think we are passing `param_set` left and right over there anyway. So the changes should be easy.

PR #100, opened 2022-05-20
Parameter Box integration with InternalClimaParams

odunbar:


charleskawczynski:
This PR:
 - Uses the new CLIMAParameters interface
 - Drops support for Julia 1.4, since it doesn't allow some NamedTuple syntax ( I think, e.g., `nt = (; a)`)

charleskawczynski:
@odunbar, I've rebased this an I think that the PR is about ready. My only remaining concern is that maybe we should reconsider the name `param_struct` (nearly everything is a struct, so it's not really informative). Maybe we should use `param_dict`? We should probably update CLIMAParameters test / example to use this terminology, too.

docs/src/saturation_adjustment.jl
```diff
@@ -6,10 +6,11 @@ FT = Float64;
 import Thermodynamics as TD
 const ICP = TD.InternalClimaParams
 import CLIMAParameters as CP
-struct EarthParameterSet <: CP.AbstractEarthParameterSet end;
-const param_set = EarthParameterSet();
 
 TD.print_warning() = false
+param_struct = CP.create_parameter_struct(FT; dict_type = "alias")
+param_set = ICP.ThermodynamicsParameters(param_struct)
```
charleskawczynski, line 12: Should probably make this const

charleskawczynski:
Hmm, there’s a pretty significant memory alloc hit for the p-theta-a constructor.. it doesn’t seem to exist with the other constructors, we should probably look into this..

charleskawczynski:
I've also made a few more compactifying changes:

```julia
for fn in fieldnames(ThermodynamicsParameters)
    @eval $(fn)(param_set::ATP) = param_set.$(fn)
end
```
now replaces all the manual definitions. In addition, derived keywords expressions are written in the struct, and we instead compute aliases by:
```julia
    derived_params = [
        "R_d",
        "molmass_ratio",
        "R_v",
        "LH_f0",
        "e_int_v0",
        "e_int_i0",
        "cp_d",
        "cv_d",
        "cv_v",
        "cv_l",
        "cv_i",
    ]
    aliases =
        setdiff(string.(fieldnames(ThermodynamicsParameters)), derived_params)
```

I think that this may be a better pattern since there will (I think) be generally fewer derived quantities than non-derived ones. So specifying this list is a bit more compact.

Perhaps we can define a method in CLIMAParameters,`required_kwargs` that returns the symbols of all kwargs without defaults.

Issue #73, opened 2021-12-12
Add tests that constructors are zero-allocating


charleskawczynski:

charleskawczynski:
We've made some of them zero-allocating in #71, we should expand this analysis and add tests for other constructors as well.

PR #106, opened 2022-05-24
Extend JET tests

charleskawczynski:
This was missed in #105.


charleskawczynski:
Closes #73 (at least for the phase equil constructors, I suppose that we could add tests for all constructors)

PR #108, opened 2022-05-27
Disallow negative q_vap_sat

charleskawczynski:
@szy21 caught that we can have negative `q_vap_sat` under certain conditions. This can lead to `q_c` being larger than `q_tot` (`q_c = q_tot - q_v_sat`), which doesn't make sense.


szy21 (approved):

yairchn:
@charleskawczynski @szy21 @trontrytel 
to get negative `q_vap_sat ` you would need to have either `p > p_v_sat` or `1 - q_tot` do we know which on it is? 
both seem unphysical to me anyways

szy21:
It happens when p < p_v_sat. This is possible in the stratosphere where pressure is low but temperature is high.

PR #109, opened 2022-05-27
Fix q_vap_sat equation

charleskawczynski:
This PR reverts changes in #108, and adds a branch to ensure the relation (in `q_vap_saturation_from_pressure`), `R_d / R_v * (1 - q_tot) * p_v_sat / (p - p_v_sat)` is used only when its valid (when `p > p_v_sat`).

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -1087,8 +1087,12 @@ function q_vap_saturation_from_pressure(
     R_v::FT = ICP.R_v(param_set)
     R_d::FT = ICP.R_d(param_set)
     p_v_sat = saturation_vapor_pressure(param_set, phase_type, T)
-    q_v_sat = R_d / R_v * (1 - q_tot) * p_v_sat / (p - p_v_sat)
-    return max(q_v_sat, 0)
+    q_v_sat = if p - p_v_sat ≥ eps(FT)
+        R_d / R_v * (1 - q_tot) * p_v_sat / (p - p_v_sat)
+    else
+        FT(1) / eps(FT)
+    end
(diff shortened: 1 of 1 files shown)


src/relations.jl
```diff
@@ -1087,8 +1087,12 @@ function q_vap_saturation_from_pressure(
     R_v::FT = ICP.R_v(param_set)
     R_d::FT = ICP.R_d(param_set)
     p_v_sat = saturation_vapor_pressure(param_set, phase_type, T)
-    q_v_sat = R_d / R_v * (1 - q_tot) * p_v_sat / (p - p_v_sat)
-    return max(q_v_sat, 0)
+    q_v_sat = if p - p_v_sat ≥ eps(FT)
+        R_d / R_v * (1 - q_tot) * p_v_sat / (p - p_v_sat)
+    else
+        FT(1)
```
tapios, line 1093:
```suggestion
        FT(1)/eps(FT)
```
This would be better for the steam atmosphere limit (I know, not an immediate concern, but let's make life easier for people years down the line who want to do such things).

charleskawczynski: Sounds good. This constant corresponds to:
```julia
julia> Float64(Float32(1)/eps(Float32))
8.388608e6

julia> Float64(1)/eps(Float64)
4.503599627370496e15
```

tapios: That's fine. Really, any number >1 (1 + eps) would do. It's just important it exceeds 1.

tapios (approved):
LGTM

PR #112, opened 2022-06-06
add Clausius_Clapeyron_relation

yairchn:
Add the partial derivative of saturation vapor specific humidity with temperature given by the Clausius Clapeyron relation


tapios (changes requested):
Please introduce a `weighted_latent_heat` function and go through the thermodynamics code to use this consistently (we are currently defining the same quantity multiple times). (@charleskawczynski). 

Please also add a unit test that numerically computes the derivative of q_vap_sat w.r.t. temperature and compares it with this analytical formulation, to ensure consistency.

test/relations.jl
```diff
@@ -499,6 +499,16 @@ end
...
 
+        # test Clausius Clapeyron relation
+        ts = PhaseEquil_ρeq.(param_set, ρ, e_int, q_tot)
+        L_v = TD.latent_heat_vapor(param_set, ts)
+        L_s = TD.latent_heat_sublim(param_set, ts)
+        λ = TD.liquid_fraction(param_set, ts)
+        L = λ * L_v + (1 - λ) * L_s
+        T = TD.air_temperature(param_set, ts)
+        q_vap_sat = TD.vapor_specific_humidity(param_set, ts)
+        @test TD.∂q_vap_sat_∂T(param_set, ts) - L^2 * q_vap_sat / (R_v * T^2)
```
charleskawczynski, line 510:
```suggestion
        @test TD.∂q_vap_sat_∂T(param_set, ts) .≈ L^2 * q_vap_sat / (R_v * T^2)
```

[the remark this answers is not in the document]
tapios: Test this by numerically computing the derivative of `q_vap_sat`. The above is wrong anyway (L in place of L^2). It is a good example of how unit tests that do not test physical consistency but simply rewrite (possibly wrong) code can give a misleading sense of correctness.

src/relations.jl
```diff
@@ -221,6 +221,24 @@ vapor_specific_humidity(q::PhasePartition) = max(0, q.tot - q.liq - q.ice)
...
+The change in saturation vapor specific humidity with temperature given by the Clausius Clapeyron relation
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `ts` ThermodynamicState
+"""
+function ∂q_vap_sat_∂T(param_set::APS, ts::ThermodynamicState)
+    R_v::FT = ICP.R_v(param_set)
+    L_v = latent_heat_vapor(param_set, ts)
+    L_s = latent_heat_sublim(param_set, ts)
+    λ = liquid_fraction(param_set, ts)
+    L = λ * L_v + (1 - λ) * L_s
```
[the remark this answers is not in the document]
tapios: If you look at the code where the above snippet appears, you will see it's used as a step in computing the temperature-dependent latent heat (via a similar weighting of the difference in heat capacities). This needs to be unified and cleaned up.

src/relations.jl
```diff
@@ -221,6 +221,24 @@ vapor_specific_humidity(q::PhasePartition) = max(0, q.tot - q.liq - q.ice)
...
+    T = air_temperature(param_set, ts)
+    q_vap_sat = vapor_specific_humidity(param_set, ts)
+    return L * q_vap_sat / (R_v * T^2)
```
tapios, line 239:
```suggestion
    return q_vap_sat * (L / (R_v * T^2) - 1/T)
```
(See updated design doc. The second term was missing. We should use this function, once updated, in saturation adjustment iterations too.)

test/relations.jl
```diff
@@ -499,6 +499,16 @@ end
...
 
+        # test Clausius Clapeyron relation
+        ts = PhaseEquil_ρeq.(param_set, ρ, e_int, q_tot)
+        L_v = TD.latent_heat_vapor.(param_set, ts)
+        L_s = TD.latent_heat_sublim.(param_set, ts)
+        λ = TD.liquid_fraction.(param_set, ts)
+        L = λ .* L_v .+ (1 .- λ) .* L_s
+        T = TD.air_temperature.(param_set, ts)
+        q_vap_sat = TD.vapor_specific_humidity(param_set, ts)
+        @test all(TD.∂q_vap_sat_∂T(param_set, ts) .≈ L .^ 2 .* q_vap_sat ./ (R_v .* T.^2))
```
tapios, line 510: Update test with correct relation. And this isn't really a test--it just recodes the function. It would be better to test this with AD or small finite differences.

src/relations.jl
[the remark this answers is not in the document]
charleskawczynski: I think I've applied the new function, `weighted_latent_heat`, to the additional place to unify, can you confirm that's all that's needed?

charleskawczynski:
@tapios, I think this is ready to go if you approve.

Once this is merged, I'll rebase the other PRs and we should hopefully be able to close out a bunch of the open issues.

src/relations.jl
```diff
@@ -911,13 +949,13 @@ function saturation_vapor_pressure(
...
-    liquid_frac = liquid_fraction(param_set, T, phase_type, q)
-    ice_frac = 1 - liquid_frac
+    λ = liquid_fraction(param_set, T, phase_type, q)
+    ice_frac = 1 - λ
 
     # effective latent heat at T_0 and effective difference in isobaric specific
     # heats of the mixture
-    LH_0 = liquid_frac * LH_v0 + ice_frac * LH_s0
-    Δcp = liquid_frac * (cp_v - cp_l) + ice_frac * (cp_v - cp_i)
+    LH_0 = λ * LH_v0 + ice_frac * LH_s0
```
tapios, line 957: Is there an advantage of having `ice_frac` here instead of `(1 - λ)`?

test/relations.jl
```diff
@@ -516,6 +517,36 @@ end
...
+                oftype(_T, _ρ),
+                _phase_type,
+                _q_pt,
+            )
+        end
+        ∂q_vap_sat_∂T_fd = _T -> ForwardDiff.derivative(q_vap_sat, _T)
+        @test isapprox(
+            ∂q_vap_sat_∂T_fd(ts_sol.T),
+            TD.∂q_vap_sat_∂T(param_set, ts_sol);
+            atol = 6.1e-3,
```
tapios, line 547: Can we make this an `rtol`?

tapios (approved):
LGTM. Thanks!

src/relations.jl
charleskawczynski: Not really, we can inline it 🙂

test/relations.jl
charleskawczynski: We can, but ouch, this it looks like it passes at `rtol = 0.9` but fails at `rtol = 0.8`. Am I interpreting that right that it's an 80% error? I'm wondering if I did something wrong somewhere

tapios: Maybe plot both (analytical and FD) to be sure? Identifying those types of issues is what tests are for :)

charleskawczynski: Plotting showed good agreement after all, is it worth adding that plot to the docs? Or is the (log) test sufficient?

tapios: It's nice to have the plots, I think.

tapios (approved):
Looking good. Thank you, @charleskawczynski!

PR #120, opened 2022-07-28
add Virtual dry static energy

yairchn:
Add dry static energy

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -44,6 +44,7 @@ export liquid_fraction, PhasePartition_equil
 # Auxiliary functions, e.g., for diagnostic purposes
 export dry_pottemp
 export virtual_pottemp
+export virtual_dry_static_energy
 export exner
 export shum_to_mixing_ratio
 export mixing_ratios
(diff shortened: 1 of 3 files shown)


szy21:
Should we add some tests for this?

src/relations.jl
```diff
@@ -2714,6 +2715,24 @@ function moist_static_energy(
     return specific_enthalpy(param_set, ts) + e_pot
 end
 
+"""
+    virtual_dry_static_energy(param_set, ts, e_pot)
+
+Virtual dry static energy, given
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `ts` a thermodynamic state
+ - `e_pot` potential energy (e.g., gravitational) per unit mass
```
tapios, line 2724:
```suggestion
 - `e_pot` gravitational potential energy per unit mass
```
(Don't we call this Phi elsewhere?)

charleskawczynski (approved):

szy21:
@charleskawczynski What did I do wrong on the doc string?

src/relations.jl
szy21: I'm not very familiar with the thermodynamics package, but I'm following e.g. moist static energy [here](https://github.com/CliMA/Thermodynamics.jl/blob/9a4bd9d145cbfff879f6768e737276ee47951750/src/relations.jl#L2709). Maybe we call this Phi somewhere in ClimaAtmos? In that case we can make it consistent in ClimaAtmos later.

tapios: e_pot is fine if that is what we do in Thermodynamics.

Issue #119, opened 2022-07-28
Consider caching saturation vapor pressure


charleskawczynski:
Computing saturation vapor pressure includes _both_ calls to `exp` and `^`, and based on looking at flame graphs in ClimaAtmos, it seems that this is actually pretty expensive.

PR #123, opened 2022-08-08
Use p_vap_sat func in q_vap_sat

charleskawczynski:
This PR is a small refactoring step towards #119. This PR:
 - Slightly generalizes `saturation_vapor_pressure` to allow users to provide a phase partition
 - `saturation_vapor_pressure` is then called inside `q_vap_saturation`, whose contents were moved over

This mostly just deletes duplicate code. There is a bit of algebraic difference between the two, but I've confirmed that they're equivalent.

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -902,16 +902,25 @@ function saturation_vapor_pressure(
     param_set::APS,
     ::Type{phase_type},
     T::FT,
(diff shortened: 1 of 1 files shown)


charleskawczynski:
I've looked over this once more, and I realize now that we have correctness tests that cover this, so merging!

PR #128, opened 2022-09-13
Add optional initial temperature guess

### LINKED ISSUE #32: Add option to pass `T_initial_guess` to `PhaseEquil`
This will allow us to use the previous temperature as an initial guess in saturation adjustment.

charleskawczynski:
This PR:
 - Adds an optional initial temperature guess
 - Adds appropriate docs
 - Fixes some docs

@tapios, I've added a test that I think may be useful for us to review interactively.

Closes #32.

### DIFF
--- a/test/relations.jl
+++ b/test/relations.jl
@@ -1670,6 +1670,42 @@ end
     @test all(phase_type .== (nt.phase_type for nt in profiles))
 end
 
+@testset "Thermodynamics - test T_guess" begin
+    ArrayType = Array{Float64}
+    FT = eltype(ArrayType)
+    param_set = parameter_set(FT)
+    profiles = TestedProfiles.PhaseEquilProfiles(param_set, ArrayType)
+    @unpack p, ρ, e_int, h, θ_liq_ice, q_tot, T, phase_type = profiles
+    T_guess = T .+ (FT(0.2) .* randn(FT, length(T)))
+    args = (q_tot, 40, FT(1e-1))
+    ts =
+        PhaseEquil_ρeq.(param_set, ρ, e_int, args..., RS.NewtonsMethod, T_guess)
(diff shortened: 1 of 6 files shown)


test/relations.jl
```diff
@@ -1614,5 +1617,49 @@ end
...
+    ts =
+        PhaseEquil_ρeq.(
+            param_set,
+            ρ,
+            e_int,
+            q_tot,
+            8,
+            FT(1e-1),
+            RS.NewtonsMethod,
+            T .+ (FT(0.1) .- rand(FT, length(T))),
```
tapios, line 1635: D you want this to be unbiased w.r.t. to the true solution, i.e.,
```suggestion
            T .+ (FT(0.1) .* rand(FT, length(T))),
```

test/relations.jl
```diff
@@ -1614,5 +1617,49 @@ end
...
+    ts =
+        PhaseEquil_ρeq.(
+            param_set,
+            ρ,
+            e_int,
+            q_tot,
+            2,
+            FT(1e-1),
+            RS.NewtonsMethod,
+            T .+ (FT(0.5) .- rand(FT, length(T))), # provide an initial guess +/- 0.5 K from true solution
```
tapios, line 1656: As above, how about this:
```suggestion
            T .+ (FT(0.5) .* rand(FT, length(T))), # provide an initial guess +/- 0.5 K from true solution
```

tapios (approved):
LGTM. Some comments on tests below. 

Would it be difficult to set maxiter by default to 1 is T_guess if given, and to something larger otherwise? If so, we may get away with a reasonable O(0.1 K) convergence tolerance in all cases.

test/relations.jl
charleskawczynski: Ah, yes. This was a last minute typo, thanks!

PR #132, opened 2022-09-19
Add data collection mechanism

charleskawczynski:
One concern we've had recently is whether our tolerance for saturation adjustment is reasonable or not because it's a bit difficult to reason about. If the tolerance is occasionally poor, we could have convergence issues with reasonable inputs and, at the moment, it's a bit difficult to know that since the inputs are not in more intuitive variables like temperature. #128 is an attempt to make sure that we start with a reasonable guess, however, knowing whether this is a good idea or not requires collecting data from a real-world run, and we don't have a clean way to do that at the moment.

This PR adds a module dedicated to collecting data, to help us better understand statistics of some important information:
 - Maximum number of iterations performed
 - Average number of iterations performed
 - Number of converged and non-converged calls (if / when we set `TD.error_on_non_convergence() = false`)

Here's a simple script for running the moist baroclinic wave in ClimaAtmos:

```julia
using Revise; include("examples/hybrid/cli_options.jl");
dict = parsed_args_per_job_id();
parsed_args = dict["sphere_baroclinic_wave_rhoe_equilmoist"];
parsed_args["enable_threading"] = false
import Thermodynamics
import RootSolvers
Thermodynamics.solution_type() = RootSolvers.VerboseSolution()
include("examples/hybrid/driver.jl")
Thermodynamics.DataCollection.print_summary()
```

At the moment, this produces

```julia
julia> Thermodynamics.DataCollection.print_summary()
┌ Info: Thermodynamics saturation_adjustment statistics:
│   max_iter = 1
│   call_counter = 15904225
│   average_max_iter = 6.287637404526156e-8
│   converged_counter = 15904225
└   non_converged_counter = 0
```
Which seems pretty good, however, this is running at coarse resolution, for only 6 days (1152 `step!`s). Runtime was about 4 min, so we can definitely crank things up.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -136,3 +136,9 @@ TemperatureProfiles.DecayingTemperatureProfile
 ```@docs
 Thermodynamics.TestedProfiles
 ```
+
+## Data collection
+
+```@docs
+Thermodynamics.DataCollection
+```
--- a/src/DataCollection.jl
+++ b/src/DataCollection.jl
@@ -1,0 +1,77 @@
+"""
+    DataCollection
+
+This module is designed to help judge the accuracy and
+performance for a particular formulation, tolerance, and or
+solver configuration, by providing tools to collect various
+statistics when Thermodynamic `saturation_adjustment` is called.
+
+## Example:
+```
+import Thermodynamics as TD
+import RootSolvers as RS
+
+function do_work()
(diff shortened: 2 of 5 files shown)


charleskawczynski:
I think this is ready to go, and evaluating things with this merged will be easier.

Issue #135, opened 2022-09-22
Saturation adjustment enhancements


charleskawczynski:
## Purpose
@cmbengue requested an SDI be made to track several changes needed in Thermodynamics. This issue will outline them.

## Background
One concern we've had recently is whether our tolerance for saturation adjustment is reasonable or not because it's a bit difficult to reason about. If the tolerance is occasionally poor, we could have convergence issues with reasonable inputs and, at the moment, it's a bit difficult to know that since the inputs are not in more intuitive variables like temperature. #128 is an attempt to make sure that we start with a reasonable guess, however, knowing whether this is a good idea or not requires collecting data from a real-world run, and we don't have a clean way to do that at the moment. (msg copied from #132)

Some planned enhancements include:
 - Adding a data collection mechanism to help compare equations/formulations and solution method parameters (maximum number of iterations allowed, initial guesses, solution tolerance etc.)
 - Improve the accuracy, and hopefully convergence of the Clausius Clapeyron relation (#112)
 - Change tolerance to something more physically intuitive. There are two options here:
     - Change `ResidualTolerance` to `SolutionTolerance`
     - Change `e_tot(T) - e_tot_sat(T)` to `air_temperature(e_tot(T)) - air_temperature(e_tot_sat(T))` and still use `ResidualTolerance`. I tried this out locally and it does work (i.e., converges and passes tests)
 - One benefit I see with the latter is that we'll still be solving for `f(x) ≈ 0`, rather than solving for when `Δx_k < tol` for the `k`-th iteration.
 - Collect data from a real world run
     - Use the moist baro wave as an example
     - Run until clouds exist--when `(q_tot+q_ice) > threshold` (where `threshold ~ 0.01`? cc @tapios)

## Cost/Benefits/Risks
(Analyze the cost/benefits/risks associated with the proposal)

One question I have is: do we want to merge #132, or should we just use it during data collection and close the PR afterwards? We'll never want to use the data collection in production runs, I think. I don't have a strong preference either way.

I suppose that if we want to compare many different things, e.g., different `T_guess` estimates, different formulations, that require Thermodynamics code changes, it may be easier if it's part of the main branch. Maybe merging is just more practical

## People and Personnel
- Lead: @charleskawczynski
- Collaborators: @yairchn, @tapios
- Reviewers: @yairchn, @tapios

## Components
Components are in the mentioned PRs

## Inputs
Not applicable?

## Results and deliverables
[...]

Issue #134, opened 2022-09-21
Several Thermodynamics updates.


tapios:
A few updates to Thermodynamics and the documentation:

- [ ] Update the thermodynamics documentation following the tracked changes in the design doc.
- [ ] Update the code according to tracked changes in design doc. Most importantly, add the missing `1/T` term in `∂q_vap_sat_∂T` in [l. 1402](https://github.com/CliMA/Thermodynamics.jl/blob/20a3abfcf930eb4e4ba2d586f34e6739149e9703/src/relations.jl#L1402). (This will change the gradient by about 5%; thus, it may lead to a minor improvement in convergence of saturation adjustment, but the changes should not be large.)
- [ ] Change the saturation adjustment convergence check generally (in all version of SA) to a solution (i.e., temperature) tolerance, requiring temperature changes `ΔT` from iteration to iteration to be small in relative terms, i.e., `ΔT/T < 1e-4`.
- [ ] Enable passing an initial state for saturation adjustment 
- [ ] Check average number of iterations needed with previous time step (or stage) as initial state and `ΔT/T < 1e-4`. Compare that exist the current default initial state. (Expectation is that fewer iterations are needed with previous time step as initial state.)
- [ ] If previous test yields expected results (faster convergence with previous timestep as initial state), make that the default in ClimaAtmos.

Some other minor fixes:
- [ ] Use `liquid_frac` and  `λ` consistently. E.g., replace `liquid_frac` in ll. [914 ff.](https://github.com/CliMA/Thermodynamics.jl/blob/20a3abfcf930eb4e4ba2d586f34e6739149e9703/src/relations.jl#L914) and [ll. 1314 ff.](https://github.com/CliMA/Thermodynamics.jl/blob/20a3abfcf930eb4e4ba2d586f34e6739149e9703/src/relations.jl#L1301) by  `λ`
- [ ] Replace `specific entropy` in [l. 2846](https://github.com/CliMA/Thermodynamics.jl/blob/20a3abfcf930eb4e4ba2d586f34e6739149e9703/src/relations.jl#L2846) and [l. 2865](https://github.com/CliMA/Thermodynamics.jl/blob/20a3abfcf930eb4e4ba2d586f34e6739149e9703/src/relations.jl#L2865) by `partial pressure`

charleskawczynski:
[...]

PR #137, opened 2022-09-28
Change tolerance from energy to temperature

### LINKED ISSUE #130: Use SolutionTolerance for SA tolerance
Deriving a meaningful `ResidualTolerance` for a variety of inputs is bit difficult, a simpler solution is to change to using SolutionTolerance, since the magnitude and criteria is clear

charleskawczynski:
This PR changes the tolerance from energy to temperature by:
 - Wraps the left and right-hand sides of the root equation in a call to `air_temperature`, so that the residual has units of temperature
 - Then changes the ResidualTolerance from energy to temperature

This has the benefit of using a more direct tolerance of what we care about (the range on temperature).

I need to apply this to the other SA methods that use energy

Closes #130. A step towards #134 and #135.

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -1489,7 +1489,7 @@ function saturation_adjustment(
     _T_min::FT = TP.T_min(param_set)
     cv_d::FT = TP.cv_d(param_set)
     # Convert temperature tolerance to a convergence criterion on internal energy residuals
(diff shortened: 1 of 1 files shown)


tapios (approved):
LGTM

charleskawczynski:
> LGTM

Ah, sorry I forgot to close this. Superseded by #139.

PR #140, opened 2022-10-11
Use RelativeSolutionTolerance

charleskawczynski:
This PR addresses one of the check boxes in #134, by changing from `SolutionTolerance` to `RelativeSolutionTolerance`.

One downside of this, unfortunately, is that it's a breaking change, but I think it's fine if we know that this is what we want in the long run.

Alternatively, we could require users to pass in `::RootSolvers.AbstractTolerance`, which would make the interface less likely to be breaking, however this means that our interface would be tied to using `RootSolvers`, so it's not a complete win to do that.

It looks like all the tests pass with these changes.

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -1450,7 +1450,7 @@ end
         q_tot,
         phase_type,
         maxiter,
-        temperature_tol,
+        relative_temperature_tol,
         T_guess,
     )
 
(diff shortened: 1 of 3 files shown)


charleskawczynski:
Note to self: try without catching `T_freezing` edge cases

PR #141, opened 2022-10-18
Change to RootSolvers tolerance

charleskawczynski:
This is an alternative to #140.

Pros: the interface is likely more stable (assuming we'll continue with RootSolvers), since this gives us a lot of flexibility in specifying the tolerance.
Cons: users must depend on RootSolvers if they want to use custom tolerances

Similar to #140, in this PR we don't directly have the `temperature_tol`, and need some estimate for handling how we break out into freezing edge cases:

```julia
freezing_temperature_tol(tol::RS.ResidualTolerance, ::Type{FT}) where {FT} =
    tol.tol
freezing_temperature_tol(tol::RS.AbstractTolerance, ::Type{FT}) where {FT} =
    FT(0.1)

# then, when we need temperature_tol:
temperature_tol = freezing_temperature_tol(tol, FT)
```

### DIFF
--- a/test/relations.jl
+++ b/test/relations.jl
@@ -608,7 +608,7 @@ end
         @test all(air_temperature.(param_set, ts) .== Ref(_T_freeze))
 
         # PhaseEquil
-        ts_exact = PhaseEquil_ρeq.(param_set, ρ, e_int, q_tot, 100, FT(1e-3))
+        ts_exact = PhaseEquil_ρeq.(param_set, ρ, e_int, q_tot, 100, FT(1e-6))
         ts = PhaseEquil_ρeq.(param_set, ρ, e_int, q_tot)
         @test all(
             isapprox.(
@@ -683,7 +683,7 @@ end
                 e_int,
                 q_tot,
                 100,
-                FT(1e-3),
+                FT(1e-6),
                 RS.SecantMethod,
             )
         ts =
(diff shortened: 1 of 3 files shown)


test/relations.jl
```diff
@@ -602,13 +602,21 @@ end
...
         # PhaseEquil
-        ts_exact = PhaseEquil_ρeq.(param_set, ρ, e_int, q_tot, 100, FT(1e-3))
+        ts_exact =
+            PhaseEquil_ρeq.(
+                param_set,
+                ρ,
+                e_int,
+                q_tot,
+                100,
+                RS.RelativeSolutionTolerance(FT(1e-6)),
```
tapios, line 618: I'd prefer this not to be hardwired. E.g., use `sqrt(eps)` in place of `1e-6`.

test/relations.jl
```diff
@@ -683,7 +691,7 @@ end
                 e_int,
                 q_tot,
                 100,
-                FT(1e-3),
+                RS.RelativeSolutionTolerance(FT(1e-6)),
```
tapios, line 694: Ditto.

tapios (approved):
This looks fine. But I'm not sure we need this flexibility. 

An added advantage of just using solution tolerances likely is that with those, we do not need to catch the non-differentiable region around freezing. We just needed to do that if we specify tolerances on energy variables, which jump at the phase transition. But temperature is continuous, so with solution tolerances everything probably works without the edge case catch.

charleskawczynski:
Superseded by #140

PR #151, opened 2023-08-28
Dont assume that MSLP and theta ref pressure are the same

trontrytel:
We were assuming that mean sea level pressure (MSLP) and the reference pressure used in potential temperature are the same. This is technically not a mistake. It is causing problems in ClimaAtmos where we were overwriting `MSLP=101325` value to be `100000` hPa for some of our cases but not for others.

I'm hoping that differentiating between the two would help us get rid of those [overwrites](https://github.com/CliMA/ClimaAtmos.jl/blob/d91fe26e69a964746fea7778d98ab8ab760ed598/src/parameters/create_parameters.jl#L133-L156)

I kept the `MSLP` in temperature profiles and in the `specific_entropy_` reference pressure. Lucky for us entropy is not used in ClimaAtmos, so it doesn't really matter.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Thermodynamics"
 uuid = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 authors = ["Climate Modeling Alliance"]
-version = "0.10.2"
+version = "0.11.0"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
--- a/src/Parameters.jl
+++ b/src/Parameters.jl
@@ -21,6 +21,7 @@ param_set = TP.ThermodynamicsParameters{FT}(; param_pairs...);
 Base.@kwdef struct ThermodynamicsParameters{FT}
     T_0::FT
     MSLP::FT
+    p_ref_theta::FT
     cp_v::FT
     cp_l::FT
     cp_i::FT
--- a/src/isentropic.jl
+++ b/src/isentropic.jl
@@ -28,10 +28,10 @@ function air_pressure_given_θ(
     Φ::FT,
     ::DryAdiabaticProcess,
 ) where {FT <: AbstractFloat}
-    MSLP::FT = TP.MSLP(param_set)
+    p0::FT = TP.p_ref_theta(param_set)
     _R_d::FT = TP.R_d(param_set)
     _cp_d::FT = TP.cp_d(param_set)
-    return MSLP * (1 - Φ / (θ * _cp_d))^(_cp_d / _R_d)
+    return p0 * (1 - Φ / (θ * _cp_d))^(_cp_d / _R_d)
 end
 
 """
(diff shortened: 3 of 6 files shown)


trontrytel:
How set in stone is this number in the test?

```
  Expression: count(air_temperature.(param_set, ts_mid) .== Ref(_T_freeze)) ≥ 1373
   Evaluated: 1296 ≥ 1373
```

src/TemperatureProfiles.jl
```diff
@@ -77,14 +77,14 @@ function (profile::DryAdiabaticProfile)(param_set::APS, z::FT) where {FT}
...
-    MSLP::FT = TP.MSLP(param_set)
+    p0::FT = TP.p_ref_theta(param_set)
 
     # Temperature
     Γ = grav / cp_d
     T = max(profile.T_surface - Γ * z, profile.T_min_ref)
 
     # Pressure
-    p = MSLP * (T / profile.T_surface)^(grav / (R_d * Γ))
+    p = p0 * (T / profile.T_surface)^(grav / (R_d * Γ))
```
szy21, line 87: We probably don't need to change this.

trontrytel: Fixed

charleskawczynski:
> How set in stone is this number in the test?

Yeah, that was a bit of a kludge. We should really be testing this: https://github.com/CliMA/Thermodynamics.jl/blob/main/test/relations.jl#L837. I forget exactly why that wasn’t working.

tapios (approved):
This looks good. It's good to separate the different reference pressures playing different roles in thermodynamic functions. 

Given that MSLP is only used as the reference pressure for entropies now, could you please rename this to `p_ref_entropy` or something like it, to make clear what role it plays?

charleskawczynski (approved):
My only concern is, is it possible that this opens the door for inconsistencies? Otherwise it seems good.

charleskawczynski (commented):
Also, I agree that this is a breaking change, so will need to bump the minor version

tapios:
> > Given that MSLP is only used as the reference pressure for entropies now, could you please rename this to `p_ref_entropy` or something like it, to make clear what role it plays?
> 
> @tapios - I changed the name in `Thermodynamics` but kept the `MSLP` parameter in `CLIMAParameters`. Or would you rather I also change the name in `CLIMAParameters`?
> 
> We use `MSLP` in a couple of places in `ClimaAtmos`. In some of those places it should now be replaced with `p_ref_theta`. I wanted to do that after I update `Thermodynamics`.
> 
> We use `MSLP` in `ClimaAtmos` to:
> 
> * to compute the reference density: `ᶜρ_ref = @. MSLP * exp(-grav * ᶜcoord.z / (R_d * T_ref)) / (R_d * T_ref)` (don't have to be changed)
> * in initial profiles (I think in some of those it should be changed to `ref_p_theta`, but not all)
> * in HeldSuarez forcing (I think it should be changed for theta forcing, but not needed for energy)
> * in implicit tendency wfact for theta (should be changed to `p_ref_theta`)

Of course, please do leave MSLP as a parameter if it's used for initial conditions etc. in ClimaAtmos. Please make the changes to `p_ref_theta` in ClimaAtmos where needed to disambiguate the parameters.

trontrytel:
bors try

charleskawczynski:
Ah, right, we switched to the GH merge queue

charleskawczynski:
Let's try

PR #155, opened 2023-09-01
Update RootSolvers, avoid recomputation

charleskawczynski:
This PR updates to the latest RootSolvers and avoids recomputing some variables during saturation adjustment.


trontrytel (approved):
LGTM! Curious to see the impact of it on the simulation times

PR #158, opened 2023-09-02
Introduce logger type, remove module methods

charleskawczynski:
This PR removes `print_warning` and `error_on_non_convergence`, and retains the functionality by introducing logger types.


trontrytel:
Is that something I could also use in CloudMicrophysics to print warnings when someone is trying to use the parameterizations outside of the acceptable range of inputs? 

I don't want to throw an error. But it would be great to have a warning of some sort

simonbyrne:
@vchuravy this is where I wanted some sort of GPU logger.

I'm not 100% sure of names. I would prefer if it didn't get conflated with the existing Julia logging though.

charleskawczynski:
It seems like we may no longer have a choice in CUDA 5.0: https://buildkite.com/clima/climaatmos-ci/builds/13625#018b0189-731b-4c9d-970f-554d7c5740b1/145-235. Going to move forward with this. We don't need to actually use this, but it will nevertheless help with upstream inference.

charleskawczynski:
@simonbyrne, it seems that this functionality will help us get https://github.com/CliMA/ClimaAtmos.jl/pull/2199 merged in. Unless there are naming objections/suggestions, I'd like to merge this.

Sbozzolo:
Can you please give a brief high-level overview on how this works?

charleskawczynski:
Yes. Although, I keep having back-and-forth thoughts on this.

One downside that I'm remembering is that: The src code (e.g., in ClimaAtmos) will need to be changed if we leverage this feature to silence JET noise. Right now, we can just use `TD. print_warning() = false` at the top of the JET script.

Maybe we can put this on hold.

PR #165, opened 2023-09-06
Change supersaturation definition

trontrytel:
It actually shouldn't change anything right now. But i still think it will simplify life for us in CloudMicrophysics. - `S` is typically defined as `p_vap / p_vap_sat` . Different papers provide different parameterizations for `p_vap_sat` and it would be good to test their impact on S and ice formation without having to go through recomputing it to `q_vap_sat`

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -1154,34 +1154,34 @@ end
  - `Liquid()`, `Ice()` - liquid or ice phase to dispatch over.
  - `ts` thermodynamic state
 
-Returns supersaturation (qv/qv_sat -1) over water or ice.
+Returns supersaturation (pv/pv_sat -1) over water or ice.
 """
 function supersaturation(
     param_set::APS,
     q::PhasePartition{FT},
     ρ::FT,
     T::FT,
-    ::Liquid,
+    phase::Phase,
 ) where {FT <: Real}
 
-    q_sat::FT = q_vap_saturation_generic(param_set, T, ρ, Liquid())
-    q_vap::FT = vapor_specific_humidity(q)
+    p_v_sat::FT = saturation_vapor_pressure(param_set, T, phase)
+    p_v::FT = vapor_specific_humidity(q) * (ρ * TP.R_v(param_set) * T)
 
-    return q_vap / q_sat - FT(1)
+    return p_v / p_v_sat - FT(1)
 end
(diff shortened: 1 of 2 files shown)


src/relations.jl
```diff
@@ -1154,33 +1154,20 @@ end
...
-    q::PhasePartition{FT},
-    ρ::FT,
-    T::FT,
-    ::Ice,
+    phase::Phase,
 ) where {FT <: Real}
 
-    q_sat::FT = q_vap_saturation_generic(param_set, T, ρ, Ice())
-    q_vap::FT = vapor_specific_humidity(q)
+    p_v_sat::FT = saturation_vapor_pressure(param_set, T, phase)
```
charleskawczynski, line 1167: It's difficult for me to use the GH suggestion, but let's make `p_v_sat` an optional argument (since it's relatively expensive to compute):

```julia
    ...
    phase::Phase,
    p_v_sat::FT = saturation_vapor_pressure(param_set, T, phase)
) where {FT <: Real}

   ...
```

trontrytel: So you want one version of `supersaturation` where `p_v_sat` is provided as input and another when it's computed from `T`?

charleskawczynski: I think we can just move the `p_v_sat::FT = ...` up a few lines to make a single method with an optional argument.

trontrytel: So we would no longer support `supersaturation(param_set, ts, phase)`. And instead support `supersaturation(p, p_sat)` ?

charleskawczynski: The change I'm suggesting is non-breaking compared to the original PR, I just want `p_v_sat` to be an optional argument in `supersaturation`. `supersaturation(::ThermoParams, ::ThermoState)` can still call that method, and it doesn't need to pass the optional argument-- it's just helpful to have that argument optional in case we call this function from somewhere that `p_v_sat` has already been computed

trontrytel: Gihub got really confused with the changes, it's better to just take a look at the file. 

I now have `supersaturation(..., phase)` and `supersaturation(..., p_v_sat)` The user facing method `supersaturation(params, ts, phase` goes back to `supersaturation(..., phase)`

charleskawczynski (approved):

PR #171, opened 2023-12-02
Improve the explanation surrounding "virtual temperature" and a hypothetical "density temperature"

glwagner:
I thought the description could be improved by being a little more explicit. If I've made a mistake, let me know and I can fix it. I also embellished the "side point" by putting it in a box to emphasize that this is additional information that isn't core to understanding the main thread.

PR #172, opened 2023-12-05
Add capability for user-defined parameter sets, with an example

glwagner:
This PR tweaks the internal API for defining parameter sets to support user-defined parameters. This is important, because it will allow users to invoke functions in `Thermodynamics.jl` without requiring the full suite of parameters in `ThermodynamicsParameters.jl`. For example, if one is only interested in computing density as a function of temperature, pressure, and (total) specific humidity, then they only require the universal gas constant and the molar masses of air and water.

Next, this PR adds an example which illustrates such a parameter set, called `ConstitutiveParameters`. It culminates in this plot:

image

which uses data derived from the JRA55 reanalysis (subsampled to make the dataset smaller), showing the primary dependence of density on temperature, secondary dependence on pressure, and teritiary dependence on specific humidity for the typical conditions of the Earth's atmosphere.

Currently, there are no examples, so this PR introduces the first one. I'd be happy to turn the example into something that appears in the documentation via Literate (the example is very close to that form already).


glwagner:
I guess this PR depends on #171, but that can be removed.

docs/src/Formulation.md
```diff
@@ -356,24 +356,31 @@ where ``\Pi`` is known as the Exner function
...
+
 ```math
 \begin{equation}
 θ_v = \frac{R_m}{R_d} θ.
 \label{e:virtual_pottemp}
 \end{equation}
 ```
-Some texts distinguish virtual and density (potential) temperatures, where density (potential) temperatures take the mass of condensate into account, whereas virtual (potential) temperatures do not. We always take the mass of any condensate into account in the thermodynamics of moist air and do not make this distinction here.
+
+!!! note
```
navidcy, line 379:
```julia
!!! note "A title of the note"
```
?

examples/density_from_temperature_pressure_humidity.jl
navidcy, line 1: why don't we literate it and add it in the docs?

docs/src/Formulation.md
glwagner: good idea

glwagner: except this change is really on #171, i accidentally combined the two

examples/density_from_temperature_pressure_humidity.jl
glwagner: That's what I asked about in the original post

examples/density_from_temperature_pressure_humidity.jl
```diff
@@ -0,0 +1,155 @@
...
+```math
+Rᵐ(q) = 
+```
+
+where 
+
+For more information see [reference docs].
+"""
+function ConstitutiveParameters(FT = Float64;
+                                gas_constant       = 8.3144598,
```
trontrytel, line 45: I had some discussions a while back with SEs asking if the parameters constructors should come with default values. The feedback I got was leaning towards not having the defaults, unless it is really beneficial to the structure of the code. 

From what I understood the defaults should live in a toml file. - I don't have any opinions on this personally. More of a question of convention and how much we want to stick to it in different examples

trontrytel (approved):
+1 on turning it into a documentation page

examples/density_from_temperature_pressure_humidity.jl
glwagner: This is a "classroom example" that illustrates some thermodynamic calculations. I think we want to include examples like this in the docs as well to explain the theromdynamic theory that this package implements. This example does not illustrate how to use this package within an Earth system model, necessarily.

charleskawczynski:
@glwagner, can you re-open this on branch off the main branch? (not the fork), I wasn't able to rebase it

glwagner:
I can do that.

glwagner:
@charleskawczynski I think we just need to run the formatter

examples/density_from_temperature_pressure_humidity.jl
```diff
@@ -138,18 +142,45 @@ Tmax = maximum(T)
...
+
+scatter!(
+    axρ,
+    T[:],
+    ρ[:],
+    color = p[:],
+    colorrange = prange,
+    colormap = pmap,
+    alpha = 0.1,
+)
```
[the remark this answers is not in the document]
glwagner: I want to clean this up if possible within the rules that the formatter dictates (perhaps its not possible)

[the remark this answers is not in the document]
navidcy: This might be relevant?

https://domluna.github.io/JuliaFormatter.jl/stable/#join_lines_based_on_source

glwagner:
@charleskawczynski looks like the buildkite isn't running even though I added the "Launch buildkite" label

charleskawczynski:
> @charleskawczynski looks like the buildkite isn't running even though I added the "Launch buildkite" label

Let me try. If it works, I'll see if I can update your privileges.

charleskawczynski:
@glwagner, maybe because it's a PR from a fork? Ah, or the branch needs to be rebased (so that it picks up the latest buildkite config).

examples/density_from_temperature_pressure_humidity.jl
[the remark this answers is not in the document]
charleskawczynski: Sometimes I find this very handy

PR #176, opened 2024-01-16
Move parameter constructor to a package extension, use name maps

nefrathenrici:
This PR adds a package extension which includes new convenience constructors for `ThermodynamicsParameters` when CLIMAParameters is loaded.

The constructors are 
- `ThermodynamicsParameters(FT)`
- `ThermodynamicsParameters(toml_dict)`

The core changes are in `ext/CreateParametersExt.jl` and `Project.toml`, but I have updated the repo to work with the new parameter constructor.

This PR updates to the newest release of ClimaParameters and incorporates breaking changes.

This PR remove all instances of the `parameter_set` function and replaces it with the new constructors. 
I will also fix up the import statements if people like the core changes of this PR.

### DIFF
--- a/ext/CreateParametersExt.jl
+++ b/ext/CreateParametersExt.jl
@@ -1,0 +1,42 @@
+module CreateParametersExt
+
+import Thermodynamics.Parameters.ThermodynamicsParameters
+import CLIMAParameters as CP
+
+ThermodynamicsParameters(::Type{FT}) where {FT <: AbstractFloat} =
+    ThermodynamicsParameters(CP.create_toml_dict(FT))
+
+function ThermodynamicsParameters(toml_dict::CP.AbstractTOMLDict)
+    name_map = (;
+        :temperature_min_at_reference => :T_min_ref,
+        :entropy_water_vapor => :entropy_water_vapor,
+        :entropy_dry_air => :entropy_dry_air,
+        :potential_temperature_reference_pressure => :p_ref_theta,
+        :entropy_reference_temperature => :entropy_reference_temperature,
+        :temperature_saturation_adjustment_max => :T_max,
+        :molar_mass_dry_air => :molmass_dryair,
+        :pow_icenuc => :pow_icenuc,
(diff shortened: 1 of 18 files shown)


ext/CreateParametersExt.jl
```diff
@@ -0,0 +1,41 @@
+module CreateParametersExt
+
+import Thermodynamics.Parameters.ThermodynamicsParameters
+import CLIMAParameters as CP
+
+ThermodynamicsParameters(FT) = ThermodynamicsParameters(CP.create_toml_dict(FT))
+
+function ThermodynamicsParameters(toml_dict::CP.AbstractTOMLDict)
+    name_map = Dict{String, String}(
```
glwagner, line 9: Out of curiosity, why do we use strings instead of symbols? It seems that strings include names that should be invalid, whereas symbols are a better map to the concept of a "variable name". This would make using named tuples more natural as well rather than dictionaries (in this case, it seems we should use `NamedTuple` rather than `Dict` since we don't instead to add to or change the contents of `name_map`).

nefrathenrici: Both strings and symbols are currently supported! I am happy to turn the name map into a NamedTuple, but I will have to add support for it.

nefrathenrici:
bors try

ext/CreateParametersExt.jl
glwagner: I like `NamedTuple` because it seems a little more explicit / purpose-built for this (using `Dict` implies we need to manipulate its elements, `pop!`ing, adding, etc). `NamedTuple` evokes a clearer vision of the code's purpose, making it easier to understand. Not every optimization is worth the effort, so it's up to you.

nefrathenrici: Thanks for reviewing! I think NamedTuple is more intuitive for people following this pattern, so I will used NamedTuples as the default name map examples.

charleskawczynski (approved):

Issue #177, opened 2024-01-17
`if`-statements may reduce performance on GPUs


glwagner:
`if` statements appear throughout this package. For example:

https://github.com/CliMA/Thermodynamics.jl/blob/d73a83404eccbf5351722777dfa2f59800e5d106/src/relations.jl#L1290-L1311

In all cases except those in which one branch cannot be executed in some situation, `if` statements should be replaced by `ifelse`. For example, [CUDA.jl's "performance tips"](https://cuda.juliagpu.org/stable/tutorials/performance/#General-Tips) states:

> * Avoid using control flow instructions such as if which cause branches, e.g. replace an if with an ifelse if possible;

Note that this performance tip is not limited to GPUs; it's just that CPUs are better at branch prediction (and are slower) so the slowdown incurred by the presence of an `if` statement within a hot inner loop may be more difficult to observe.

As a rule, we should always use `ifelse`, `|` and `&` in performance-critical code, and avoid the "short-circuiting" operators `if`, `||` and `&&`.

We could expend some effort to attempt to determine whether or not the `if` statements are incurring a slowdown. However, I think this is a waste of time unless we cannot use `ifelse` for some reason. This rule of thumb should also be followed for all future code.

PR #178, opened 2024-01-18
Use `ifelse` in liquid_fraction, plus a comment

glwagner:
This PR makes moves towards resolving #177 by updating `liquid_fraction` to use `ifelse` rather than `if`, `else`.

I also changed the formatting a bit and added a comment about the saturation adjustment model that's implemented. The philosophy of the formatting change is to attempt to sharply distinguish between the two ways we have of writing expressions: mathematical notation (symbolic, concise), and in natural language (in this case, english). In my opinion, the current style does not use _either_ concise mathematical notation or natural language, but rather a kind of frankenstein of the two that I find difficult to read (eg verbose, but symbolic).

There are only a few more places that `if`, `else` occurs. If these changes are welcome then I'll also fix the others.

Also curious --- there are many annotations of the kind `FT <: Real` and `var::FT`. What is the motivation for these? I don't think they speed up compilation. They do serve as a sort of self-documentation for the code though. For example, many functions are diagonalized (many input types are forced to all have type `FT`), even though the function would likely be valid and work as intended if one or more of the types were an integer, for example.


src/relations.jl
```diff
@@ -1290,24 +1290,26 @@ is a function that is 1 above `T_freeze` and goes to zero below `T_icenuc`.
...
-        _pow_icenuc::FT = TP.pow_icenuc(param_set)
-        if _pow_icenuc == 1
-            return (T - _T_icenuc) / (_T_freeze - _T_icenuc)
-        else
-            return ((T - _T_icenuc) / (_T_freeze - _T_icenuc))^_pow_icenuc
-        end
-    else
-        return FT(0)
-    end
+) where State <: ThermodynamicState
```
charleskawczynski, line 1295:
```suggestion
) where {FT, State <: ThermodynamicState}
```

glwagner: Style guides usually recommend getting `FT` from an `eltype` rather than type parameter. Should we use that instead then?

glwagner: Hmm I see that we need it for `q_pt_0` so perhaps best to leave the type parameter

charleskawczynski: That would be fine, but this enforces diagonalization

glwagner:
Should we do a performance benchmark with ClimaAtmos before merging? If we're concerned about performance that's also a good reason to split this `ifelse` buisness into a few PRs rather than lumping into one.

charleskawczynski:
> Should we do a performance benchmark with ClimaAtmos before merging? If we're concerned about performance that's also a good reason to split this ifelse buisness into a few PRs rather than lumping into one.

Yeah, I'll add some kernel benchmarks to CI here, and that way we can get a better idea of optimizations without needing to test against ClimaAtmos.

glwagner:
> > Should we do a performance benchmark with ClimaAtmos before merging? If we're concerned about performance that's also a good reason to split this ifelse buisness into a few PRs rather than lumping into one.
> 
> Yeah, I'll add some kernel benchmarks to CI here, and that way we can get a better idea of optimizations without needing to test against ClimaAtmos.

I am happy to run some ClimaAtmos benchmarks before and after this change to report the results. I'd be more comfortable knowing for sure that the changes aren't detrimental (at the least).

charleskawczynski:
> If you can point me to it, I am happy to run some ClimaAtmos benchmarks before and after this change to report the results. I'd be more comfortable knowing for sure that the changes aren't detrimental (at the least).

I would probably suggest just running CI with this branch and look at a few GPU jobs (with moisture).

That said, we're in the middle of updating dependencies, so we'll need to wait until that is resolved. In the mean time, I did add some kernel benchmarks: https://github.com/CliMA/Thermodynamics.jl/blob/main/perf/kernel_bm.jl

charleskawczynski:
I went ahead and rebased

src/relations.jl
```diff
@@ -1320,22 +1320,27 @@ is a function that is 1 above `T_freeze` and goes to zero below `T_icenuc`.
...
+    Tᶠ = TP.T_freeze(param_set)   # freezing temperature
+    Tⁿ = TP.T_icenuc(param_set)   # temperature of homogeneous ice nucleation
+    α = TP.pow_icenuc(param_set) # power law partial ice nucleation parameter
+
+    # Model for cloud water condensate probabilty when temperature is below
+    # freezing (all liquid condensate), but above the temperature of homogeneous
+    # ice nucleation (all ice condensate). In it's simplest form, this is simply
+    # a number that varies between 0 and 1. For example, see figure 6 in 
+    # Hu et al., https://doi.org/10.1029/2009JD012384, JGR (2010).
+    λᵖ = ((T - Tⁿ) / (Tᶠ - Tⁿ))^α
```
charleskawczynski, line 1334: One reason that I introduced branching was to avoid calling the generic `pow` (which is quite expensive) in the case when `α==1`. I suppose we can use one more level of `ifelse` nestedness to recover that optimization.

glwagner: When we use `ifelse`, both branches are computed. Thus no savings can be realized; adding a computation for the case `α==1` will always produce a slow down. This is a generic feature of GPU computing --- typically, all computations are performed at every grid point, and `if` statements predicated on runtime values cannot be used for optimization. It is possible of course that for extremely heavy code, that branching / `if` statements can produce a performance benefit. But one should not expect to use such constructs.

In this case, `α==1` can actually be known at compile time. The case `α isa Int` is also important. To express this important performance optimization, the parameter `pow_icenuc` should be allowed to be a type other than `FT` so the optimization can be resolved at compile time.

charleskawczynski: I don't think we'll see much of any performance difference with removing the branch. I did see an improvement by including this branch on the CPU. `pow` is generally pretty expensive. Also, we previously ran into https://stackoverflow.com/questions/14687665/very-slow-stdpow-for-bases-very-close-to-1, which turned out to be extremely expensive and hence why we implemented `pow_hack`.

charleskawczynski (approved):
I'm fine with merging, the doc string is a nice improvement, thanks @glwagner!

PR #180, opened 2024-01-19
Update deps and julia version

charleskawczynski:
This PR:
 - Ditches UnPack (and, technically, drops support for some older julia versions in our test suite, but 🤷🏻)
 - Upgrades Julia in CI
 - Upgrades KernelAbstractions and our GPU tests


nefrathenrici (approved):
Thanks for removing UnPack! 
Is there a reason we are updating the GH actions now? I can make this change in other repositories as well.
We should also remove the ci 1.6 merge requirement.

charleskawczynski:
> Thanks for removing UnPack! Is there a reason we are updating the GH actions now? I can make this change in other repositories as well. We should also remove the ci 1.6 merge requirement.

I just haven't gotten around to it. I want to add some kernel benchmark tests, but that required upgrading TD to julia 1.10 for CUDA/KA compatibility. So that was kind of blocking me.

And yes, feel free to upgrade in other repos as well.

Agreed, I'll remove the 1.6 check

charleskawczynski:
Something seems wrong with macos on julia 1.10, seems unrelated to this PR.

PR #185, opened 2024-01-22
Update virtual_dry_static_energy

szy21:
Includes the reference T_0 and uses cp_d instead of cp_m in the virtual dry static energy calculation. This shouldn't make much difference, but is more consistent with other equations.

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -2915,9 +2915,10 @@ function virtual_dry_static_energy(
     ts::ThermodynamicState{FT},
     e_pot::FT,
 ) where {FT <: Real}
+    T_0::FT = TP.T_0(param_set)
+    cp_d::FT = TP.cp_d(param_set)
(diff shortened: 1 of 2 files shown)


src/relations.jl
```diff
@@ -2915,9 +2915,10 @@ function virtual_dry_static_energy(
     ts::ThermodynamicState{FT},
     e_pot::FT,
 ) where {FT <: Real}
+    T_0::FT = TP.T_0(param_set)
     T_virt = virtual_temperature(param_set, ts)
     _cp_m = cp_m(param_set, ts)
-    return _cp_m * T_virt + e_pot
```
szy21, line 2920: @tapios I know this is not important but since I am changing it anyway, should I change the cp_m to cp_d here?

tapios: Yes, please.

charleskawczynski (approved):

PR #192, opened 2024-02-09
Fix zero `q_tot` edge case

charleskawczynski:
This PR fixes a zero `q_tot` edge case:

```julia
using Revise
import Thermodynamics as TD
using CLIMAParameters # load convenience constructors

thermo_params = TD.Parameters.ThermodynamicsParameters(Float32)

e_int=Float32(-3.687397e7);
ρ=Float32(0.0027615533);
q_tot=Float32(0.0);
maxiter=100
tol=Float32(0.003)
ts = TD.PhaseEquil_ρeq(
    thermo_params,
    ρ,
    e_int,
    q_tot,
    maxiter,
    eltype(thermo_params)(tol),
)
```
fails on main, and passes in this PR.

This was found in the log here: https://buildkite.com/clima/climaatmos-ci/builds/16548. This could be a fix for at least one job breaking in https://github.com/CliMA/ClimaAtmos.jl/issues/2635.

By the way, being able to create parameters in this way is just so nice! (cc @nefrathenrici)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Thermodynamics"
 uuid = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 authors = ["Climate Modeling Alliance"]
-version = "0.11.6"
+version = "0.11.7"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -1573,7 +1573,7 @@ See also [`saturation_adjustment`](@ref).
     tol = RS.RelativeSolutionTolerance(relative_temperature_tol)
 
     T_1 = max(_T_min, air_temperature(param_set, e_int, PhasePartition(q_tot))) # Assume all vapor
-    if T_1 > _T_min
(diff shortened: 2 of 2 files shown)


trontrytel (approved):
Thank you!

szy21:
Thanks! Just want to note that if `e_int=Float32(-3.687397e7)`, something else is wrong.

charleskawczynski:
Yeah, we basically came to that conclusion during the want meeting. We’re going to apply one more update: remote the T_min max part so that we can get temperatures below 150 K.

but this does mean that something else is wrong in the RRTMGP job.

maybe it’d be helpful to also print the air temperature assuming all vapor/ice in the SA error?

szy21:
> Yeah, we basically came to that conclusion during the edmf meeting. We’re going to apply one more update: remove the T_min max part so that we can get temperatures below 150 K.
> 
> but this does mean that something else is wrong in the RRTMGP job.
> 
> maybe it’d be helpful to also print the air temperature assuming all vapor/ice in the SA error?

That could be helpful I think. The single column all-sky radiation job crashed as well. This is a much simpler case to look at. We should be able to run Float32 and Float64 side by side to see what is going on. I'll see if I can find anything.

PR #193, opened 2024-02-09
Use ClimaParameters T_init_min, with new T_min value

charleskawczynski:
Followup to https://github.com/CliMA/Thermodynamics.jl/pull/192


glwagner:
Do you want to use `ifelse` instead of `if` here?

charleskawczynski:
> Do you want to use `ifelse` instead of `if` here?

Not in this PR. I'll follow up on this if/when we fix https://github.com/CliMA/ClimaCore.jl/issues/1602

trontrytel (approved):

PR #194, opened 2024-02-09
Replace AbstractFloat with Real

charleskawczynski:
This should ensure that we can pass/promote duals through Thermodynamics

### DIFF
--- a/ext/CreateParametersExt.jl
+++ b/ext/CreateParametersExt.jl
@@ -3,7 +3,7 @@ module CreateParametersExt
 import Thermodynamics.Parameters.ThermodynamicsParameters
 import CLIMAParameters as CP
 
-ThermodynamicsParameters(::Type{FT}) where {FT <: AbstractFloat} =
+ThermodynamicsParameters(::Type{FT}) where {FT <: Real} =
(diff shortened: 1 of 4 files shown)


ext/CreateParametersExt.jl
```diff
@@ -3,7 +3,7 @@ module CreateParametersExt
 import Thermodynamics.Parameters.ThermodynamicsParameters
 import CLIMAParameters as CP
 
-ThermodynamicsParameters(::Type{FT}) where {FT <: AbstractFloat} =
+ThermodynamicsParameters(::Type{FT}) where {FT <: Real} =
```
glwagner, line 6: Why not remove the restriction altogether? You can test that parameter structs are built with the right type in ClimaAtmos, it may not help much to enforce this here.

charleskawczynski: I don’t think it makes sense to have a parameter struct with Ints.

glwagner: For sure only some types make sense, it's just you save yourself having to predict them in advance which might save you from having to write PRs like this (or eliminate the slowdown that happens when code doesn't work because of a wrong assumption like this, requiring source code update)

PR #195, opened 2024-02-12
Remove type casting of parameters

charleskawczynski:
We no longer need type casting, now that the parameters are controlled through typed structs, and not methods (which previously required type-casting for type stability).


glwagner (approved):
Easier to read!

PR #222, opened 2024-07-19
Change temperature tolerance of all PhaseEquil constructor to 1e-4

szy21:
Some small clean up, but hopefully it's an improvement.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,14 @@
 Thermodynamics.jl Release Notes
 ========================
 
+main
+-------
+- ![][badge-🤖precisionΔ] Change the tolerance of PhaseEquil constructor to 1e-4
+
+v0.12.7
+-------
+- ![][badge-🔥behavioralΔ] Change the upper limit of saturation specific humidity
+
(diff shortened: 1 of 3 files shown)


charleskawczynski (approved):
Code changes look fine to me. Could we change the title to "Relax temperature tolerances"?

Also, maybe we can update the news? (we can add a precision changes expected note). This will ultimately result in fewer NL solver iterations at the expense of slightly less accurate solutions. That's what we're aiming to do, correct?

szy21:
> Code changes look fine to me. Could we change the title to "Relax temperature tolerances"?
> 
> Also, maybe we can update the news? (we can add a precision changes expected note). This will ultimately result in fewer NL solver iterations at the expense of slightly less accurate solutions. That's what we're aiming to do, correct?

I changed the title and updated the news. I don't think we will see a difference in ClimaAtmos now as the constructor with a different tolerance is only used in some initial conditions. But we can expect changes in iterations if we further change the tolerance.

charleskawczynski (approved):

PR #224, opened 2024-09-12
Update dry air internal energy and enthalpy definitions (eq 8a and 12a)

trontrytel:
Follows what is described in the dycore paper

### DIFF
--- a/test/relations.jl
+++ b/test/relations.jl
@@ -221,7 +221,7 @@ end
     # test the wrapper for q_vap_saturation over liquid water and ice
     ρ = FT(1)
     ρu = FT[1, 2, 3]
-    ρe = FT(1100)
+    ρe = FT(1100) - _R_d * _T_0
     e_pot = FT(93)
     e_int = internal_energy(ρ, ρe, ρu, e_pot)
     q_pt = PhasePartition(FT(0.02), FT(0.002), FT(0.002))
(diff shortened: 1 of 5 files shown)


test/relations.jl
```diff
@@ -291,31 +291,32 @@ end
     T = FT(300)
     e_kin = FT(11)
     e_pot = FT(13)
-    @test air_temperature(param_set, _cv_d * (T - _T_0)) === FT(T)
+    @test air_temperature(param_set, _cv_d * (T - _T_0) - _R_d * _T_triple) ===
```
szy21, line 294: Should it be _T_0 instead of _T_triple here and elsewhere? (The numbers are exactly the same now but I think it should be _T_0 is they are not the same.)

szy21 (approved):
Thanks!

test/relations.jl
trontrytel: Ugh, yes! Thank you. It's correct in the source code. I don't know why I used the triple point in tests. Will fix!

charleskawczynski (commented):
The code changes look fine to me. Can we also update the "Thermodynamics Overview"/"Formulation" section of the docs?

Also, while the changes are breaking from the point of view of returning different numerical values, this PR isn't changing any interfaces, so it's not breaking in that sense. It may be easier to just make the change non-breaking so that we can easily propagate the changes through. In fact, doing that will allow us to easily try this out in Atmos and see how results are impacted.

trontrytel:
@charleskawczynski - Thanks for taking a look and suggesting the patch release. This will make the testing easier. 

I updated the docs, so if you are ok with all the changes we can merge this in?

PR #230, opened 2024-10-14
Protect against zero divisions in RH for cold T

trontrytel:
We are getting NaNs in RH for very low temperatures because our saturation vapor pressure is zero. For example

`TD.saturation_vapor_pressure(tps, FT(30), TD.Ice()) = 0` 
`TD.saturation_vapor_pressure(tps, FT(40), TD.Ice()) = 0` 
`TD.saturation_vapor_pressure(tps, FT(50), TD.Ice()) = 4.16f-42` 

This PR adds a small number in the denominator to prevent that. 

`eps(Float32(0)) = 1.0f-45`

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -2778,7 +2778,7 @@ vol_vapor_mixing_ratio(param_set, ts::ThermodynamicState) =
 """
     relative_humidity(param_set, T, p, phase_type, q::PhasePartition)
 
-The relative humidity, given
+The relative humidity (clipped between 0 and 1), given
  - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
  - `p` pressure
  - `phase_type` a thermodynamic state type
@@ -2796,7 +2796,7 @@ and, optionally,
     q_vap = vapor_specific_humidity(q)
     p_vap = q_vap * air_density(param_set, T, p, q) * R_v * T
     p_vap_sat = saturation_vapor_pressure(param_set, phase_type, T)
-    return p_vap / p_vap_sat
+    return max(FT(0), min(FT(1), p_vap / (p_vap_sat + eps(FT(0)))))
 end
 
 """
--- a/test/relations.jl
+++ b/test/relations.jl
@@ -499,6 +499,19 @@ end
     vmrs = vol_vapor_mixing_ratio(param_set, q)
     q_vap = vapor_specific_humidity(q)
     @test vmrs ≈ _molmass_ratio * shum_to_mixing_ratio(q_vap, q.tot)
+
+    # Relative humidity sanity checks
+    for phase_type in [PhaseDry, PhaseEquil, PhaseNonEquil]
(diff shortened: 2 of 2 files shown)


szy21 (commented):
Could we also limit relative humidity between 0 and 1, and add a test for saturation_vapor_pressure at cold temperature?

trontrytel:
> Could we also limit relative humidity between 0 and 1, and add a test for saturation_vapor_pressure at cold temperature?

I can add the test. I thought that Tapio wanted to avoid actually clipping the RH.

charleskawczynski (approved):
This looks fine to me. I'm just wondering if we know or can know what inputs could lead to this? Is it dominated by temperature?

szy21:
> > Could we also limit relative humidity between 0 and 1, and add a test for saturation_vapor_pressure at cold temperature?
> 
> I can add the test. I thought that Tapio wanted to avoid actually clipping the RH.

I think @tapios wants to clip RH to 1 (because this won't prevent that)? Let's double check.

trontrytel:
> This looks fine to me. I'm just wondering if we know or can know what inputs could lead to this? Is it dominated by temperature?

Yes, for T < 40K our parameterization for saturation vapor pressure returns zero Hence the nans

trontrytel:
Lets wait for me to add the test

tapios:
> > > Could we also limit relative humidity between 0 and 1, and add a test for saturation_vapor_pressure at cold temperature?
> > 
> > 
> > I can add the test. I thought that Tapio wanted to avoid actually clipping the RH.
> 
> I think @tapios wants to clip RH to 1 (because this won't prevent that)? Let's double check.

I think clipping to <= 1 makes sense, provided (a) we document this clearly in the doc string, and (b) there is no case where we use this for microphysics and assessing supersaturation. @trontrytel Is there any reason we'd want to diagnose supersaturation here?

trontrytel:
> I think clipping to <= 1 makes sense, provided (a) we document this clearly in the doc string, and (b) there is no case where we use this for microphysics and assessing supersaturation. @trontrytel Is there any reason we'd want to diagnose supersaturation here?

Sounds good. And no, I was just thinking that for debugging Atmos it might be easier to see the actual RH numbers we get from the division than to mask them with clipping. But we can always add s`aturation_excess` to diagnostics if we need it.

szy21 (approved):
Thanks! Just one more thing. Should we add the eps to relative humidity, or should we add it to saturation_vapor_pressure itself? I don't have a strong preference. The latter can prevent issues if we use saturation_vapor_pressure elsewhere, but we can also remember to limit it if we use it in the denominator.

trontrytel:
I'm voting for adding it to RH only. The only reason why we need to do anything is because we are dividing by it here.

trontrytel:
I'm going to try to merge it in, so that our nightly tiny AMIP can run with it. We can always revisit tomorrow, if we want to change something about the epsilon!

szy21:
Great. This issue may not show up in the nightly AMIP run now. I'll run one with 55km top later tonight, which will be more likely to have the issue.

PR #231, opened 2024-10-18
add ClimaCoupler downstream experiment

juliasloan25:
Adds a short (~300s) downstream coupled AMIP experiment using ClimaCoupler.jl. See the discussion in https://github.com/CliMA/Thermodynamics.jl/pull/226 to see why we choose to include this run.

### DIFF
--- a/.github/workflows/Downstream-ClimaCoupler.yml
+++ b/.github/workflows/Downstream-ClimaCoupler.yml
@@ -1,0 +1,40 @@
+# Runs a short (~300s) coupled AMIP simulation with the ClimaCoupler.jl package.
+# Note that the ClimaCoupler.jl test suite is also run in the "Downstream" workflow.
+name: Downstream-ClimaCoupler
+on:
+  push:
+    branches:
+      - main
+    tags: '*'
+  pull_request:
+
+# Needed to allow julia-actions/cache to delete old caches that it has created
+permissions:
+  actions: write
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


charleskawczynski:
> Adds a short (~300s) downstream coupled AMIP experiment using ClimaCoupler.jl. See the discussion in #226 to see why we choose to include this run.
> 
> Are there any other repos we should add this to?

Probably CloudMicrophysics.jl and SurfaceFluxes.jl?

.github/workflows/Downstream-ClimaCoupler.yml
```diff
@@ -0,0 +1,40 @@
...
+    name: downstream ClimaCoupler.jl
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: '1.10'
+      - uses: julia-actions/cache@v2
```
Sbozzolo, line 31:
```suggestion
    strategy:
      fail-fast: false
      matrix:
        version:
          - '1.10'
          - '1.11'
        steps:
      - uses: actions/checkout@v4
      - uses: julia-actions/setup-julia@v2
        with:
          version: ${{ matrix.version }}
      - uses: julia-actions/cache@v2
```

Sbozzolo: Even if you don't want to test on 1.11, I'd suggest having this pattern (with the matrix) so that you can easily add 1.11 later

Sbozzolo (approved):

PR #241, opened 2025-03-20
Implemented PhaseEquil_pTRH

jty53:
Implemented new method to which returns a thermodynamic state for a given pressure, temperature, and relative humidity. This method calls on PhaseEquil_pTq.

Implemented two correctness tests for this method. Additional tests for type stability are still needed. Also, I would greatly appreciate a double check of my second correctness test for PhaseEquil_pTRH.


src/states.jl
```diff
@@ -670,6 +670,35 @@ end
...
+ - `RH` relative humidity
+"""
+@inline function PhaseEquil_pTRH(
+    param_set::APS,
+    p::FT,
+    T::FT,
+    RH::FT,
+) where {FT <: Real}
+    phase_type = PhaseEquil{FT}
+    p_vap_sat = saturation_vapor_pressure(param_set, T, Liquid())
```
trontrytel, line 690: This function assumes that we only have liquid phase present. I think this version be better:
https://github.com/CliMA/Thermodynamics.jl/blob/562a3a3b197cc5df6a3aa20a0bc2d04a2ee04e8a/src/relations.jl#L968

src/states.jl
```diff
@@ -670,6 +670,35 @@ end
...
+    p_vap = RH*p_vap_sat
+    mmr = TP.molmass_ratio(param_set)
+    q_tot = p_vap*mmr/(p-(1-mmr)*p_vap)
```
trontrytel, line 693: I think this line assumes we only have water vapor and dry air present

src/states.jl
jty53: Looks good to me

src/states.jl
jty53: For my case this assumption won't be a problem, but perhaps we should generalize if we expect issues elsewhere?

src/states.jl
trontrytel: Or instead of complicating this example we could add a state that assumes we only have vapor? This could actually be helpful in other places like surface fluxes or land model. What do you think @charleskawczynski ?

src/states.jl
trontrytel: Or instead of complicating this example we could add a state that assumes we only have vapor? This could actually be helpful in other places like surface fluxes or land model. What do you think @charleskawczynski ?

charleskawczynski: Functionally, I think it'd be fine. I'm not sure what a good constructor name would be, though. Is there an easy way to generalize it? It might alleviate us needing to think of a good name

src/states.jl
trontrytel: I don't think so.

Issue #243, opened 2025-04-01
PhaseEquil_pTRH constructor is not general


trontrytel:
PR https://github.com/CliMA/Thermodynamics.jl/pull/241  added a constructor for `PhaseEquil_pTRH`.

- The constructor uses `saturation_vapor_pressure(param_set, T, Liquid())` so it is only valid for liquid clouds.
- When computing `q_tot` from `RH` it assumes no liquid water or ice are present.

I would suggest changing the name to avoid misusing it in actual clouds.

trontrytel:
cc @charleskawczynski

charleskawczynski:
@jty53, can you fix this?

szy21:
And could we change q_tot to RH [here](https://github.com/CliMA/Thermodynamics.jl/blob/f6da61ede1ab4ae4b39569f9b24291cc4b5fe6ac/test/relations.jl#L1443) as well? This is only testing types, so it doesn't matter. But using q_tot instead of RH for PhaseEquil_pTRH could cause confusion to other users.

Should we require approval to merge PRs in Thermodynamics.jl? I think it would be important to make sure things are correct for a low-level package like this. For example, other users may use `PhaseEquil_pTRH` without knowing it is only valid for liquid, and this can be risky.

jty53:
@szy21 It looks like the use of q_tot in the type stability test for `PhaseEquil_pTRH` was a mistake.
@charleskawczynski Would you prefer I make a new branch for the patch or commit it to the branch used for #241?

jty53:
I've opened [#244](https://github.com/CliMA/Thermodynamics.jl/pull/244) to address these issues. This new PR will need some  changes before merging, so I've opened the PR as a draft.

trontrytel:
Here is my 2-cents on how to fix this: https://github.com/CliMA/Thermodynamics.jl/pull/249

PR #247, opened 2025-04-16
Add promotion methods required for autodiff in ClimaAtmos

dennisYatunin:
This PR adds the minimal set of promotion methods required to differentiate the current versions of `set_implicit_precomputed_quantities!` and `implicit_tendency!` in ClimaAtmos using forward-mode automatic differentiation, as tested in CliMA/ClimaAtmos.jl#2730. The minor version number is also bumped to 0.12.11.

This is yet another followup to #199, #200, #201, and #203. I would like to reiterate that every further addition to the implicit cache and tendency in ClimaAtmos (e.g., nonequilibrium moisture, surface fluxes, etc.) will require more promotion methods to be added if we want to use autodiff. These promotion methods nullify the original purpose of the `::FT` type checks throughout this package, as they allow arbitrary combinations of `Float32`s and `Float64`s to be promoted to `Float64`s. We should consider removing these checks altogether, as they are introducing unnecessary friction to implicit solver development in ClimaAtmos.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Thermodynamics"
 uuid = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 authors = ["Climate Modeling Alliance"]
-version = "0.12.10"
+version = "0.12.11"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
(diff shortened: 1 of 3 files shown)


charleskawczynski:
@dennisYatunin, I think we were going to try to fix #243 before the next release, FYI. cc @trontrytel.

trontrytel:
I'm looking at that function right now. But in the spirit of not slowing Dennis Jacobians down, could we comment that piece of code out and do a release?

charleskawczynski:
> But in the spirit of not slowing Dennis Jacobians down, could we comment that piece of code out and do a r

Dennis already made a new release, so we're free to roll

PR #248, opened 2025-04-18
Add one more promotion method for autodiff

dennisYatunin:
A recent change to the ClimaAtmos implicit tendency for one-moment microphysics requires us to add a promotion method for `saturation_vapor_pressure` to support automatic differentiation. In addition, there was a bug in the `promote_phase_partition` function where I accidentally used `q_tot` instead of `q_ice`, which is fixed in this PR. The minor version number is also bumped to 0.12.12. This is a followup to #199, #200, #201, #203, and #247.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Thermodynamics"
 uuid = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 authors = ["Climate Modeling Alliance"]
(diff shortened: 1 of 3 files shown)


trontrytel (approved):

dennisYatunin:
@trontrytel says she will open a PR to fix #243 sometime soon, so I will wait for that to be merged in before making a new release.

PR #249, opened 2025-04-19
RH based state constructor

trontrytel:
Below are my fixes to the state constructor based on relative humidity:

- changed the name to `PhaseEquil_pTRH_unsat_liquid_only` and added asserts to highlight that it's only working if `T>0C` and `RH <=1`
- fixed a typo with `mollmass_ratio` (but double check my equations to be sure)
- fixed the value in test (it's different because of the above typo)
- removed this constructor from the tests that are based on atmospheric profiles because the `T>0C` and `RH<=1` are not met there.

Here are my notes:
PXL_20250419_001856303

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -31,7 +31,6 @@ PhaseEquil
 PhaseEquil_ρeq
 PhaseEquil_ρTq
 PhaseEquil_pTq
-PhaseEquil_pTRH
 PhaseEquil_pθq
 PhaseEquil_peq
 PhaseEquil_phq
@@ -106,6 +105,7 @@ vapor_specific_humidity
 virtual_pottemp
 virtual_temperature
 specific_entropy
+q_vap_from_RH_liquid
 ```
 
 ## Internal methods
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -60,6 +60,7 @@ export total_specific_enthalpy
 export moist_static_energy
 export specific_entropy
 export saturated
+export q_vap_from_RH_liquid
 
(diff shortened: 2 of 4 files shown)


trontrytel:
I don't know what is the use case for this @jty53 But maybe an easier alternative would be to add a function that computes `q_vap` from `p`, `T` and `RH` and add all the asserts there. And once you have `q_vap` just use the `PhaseEquil_pTq,` constructor.

charleskawczynski (approved):
I'm not thrilled about the name, but if this constructor really only makes sense for unsat / liquid-only, then it's perhaps safer to use a name that warns users of this edge-case assumption.

I think @dennisYatunin will still want the `promote` method:

```julia
PhaseEquil_pTRH_unsat_liquid_only(param_set::APS, p, T, q_tot) =
    PhaseEquil_pTRH_unsat_liquid_only(param_set, promote(p, T, q_tot)...)
```

And we should update the api doc string to `PhaseEquil_pTRH_unsat_liquid_only`. Aside from that, this looks fine on the software side (would be nice to get a second set of eyes to review the physics side).

jty53:
@dennisYatunin that should be fine. I am using this to instantiate Larcform1 in ClimaAtmos.jl using the initial conditions from Pithan's 2016 paper: 
image
This provides a profile of RH up to 300hPa, after which constant specific humidity q is assumed.

trontrytel:
Even more reason to just use a function to compute the q_v based on RH instead of adding a non-generic constructor of state. The assert for T > T_freeze is not needed since RH is specified with respect to water

PR #250, opened 2025-04-19
Try fixing ci 1.7 - macOS-latest

trontrytel:
Per https://discourse.julialang.org/t/how-to-fix-github-actions-ci-failures-with-julia-1-6-or-1-7-on-macos-latest-and-macos-14/117019

Older versions of Julia (such as Julia 1.6 and 1.7) do not have native binaries for Apple Silicon macOS, which is what we are given when asked for `macos-latest`. There are a couple of solutions provided in that thread. It' trying one that looks the least complicated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version:
+        julia-version:
           - '1.7'
           - '1.8'
           - '1.9'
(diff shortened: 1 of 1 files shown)


trontrytel:
Looks like this excluded too much...

.github/workflows/ci.yml
```diff
@@ -30,6 +30,12 @@ jobs:
           - windows-latest
         arch:
           - x64
+        exclude:
+          - github-runner: macos-latest # Apple Silicon
+            julia-version: '1.7'
```
charleskawczynski, line 35:
```suggestion
            version: '1.7'
```
Does this need to match the `version` above (same with below)?

trontrytel:
I tried with both `version` and `julia-version` everywhere. I don't understand why this doesn't work.

PR #254, opened 2025-06-04
Generalize cp_m

oalcabes:
Generalize cp_m function to take in qt, ql, and qi without the need for PhasePartition. Keeps PhasePartition compatibility.

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -269,17 +269,27 @@ end
 
 The isobaric specific heat capacity of moist air given
  - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
- - `q` [`PhasePartition`](@ref).
+ - `qₜ` total specific humidity of water
+ - `qₗ` specific humidity of liquid
+ - `qᵢ` specific humidity of ice
 """
-@inline function cp_m(param_set::APS, q::PhasePartition{FT}) where {FT <: Real}
+@inline function cp_m(param_set::APS, qₜ::FT, qₗ::FT, qᵢ::FT) where {FT <: Real}
     cp_d = TP.cp_d(param_set)
     cp_v = TP.cp_v(param_set)
     cp_l = TP.cp_l(param_set)
     cp_i = TP.cp_i(param_set)
-    return cp_d +
-           (cp_v - cp_d) * q.tot +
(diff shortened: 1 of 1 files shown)


src/relations.jl
```diff
@@ -269,17 +269,27 @@ end
...
+    return cp_d + (cp_v - cp_d) * qₜ + (cp_l - cp_v) * qₗ + (cp_i - cp_v) * qᵢ
+end
+
+"""
+    cp_m(param_set, q::PhasePartition)
+
+The isobaric specific heat capacity of moist air given
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `q` [`PhasePartition`](@ref).
+"""
```
charleskawczynski, line 290:
```suggestion
```
I think we can just add the `cp_m(param_set::APS, qₜ::FT, qₗ::FT, qᵢ::FT)` signature to the doc string above, since they're two methods for the same function.

charleskawczynski (commented):
This seems fine to me, but why do we want to change this internal method?

src/relations.jl
charleskawczynski: Oh, yeah, I guess we're not really consistent with it. A pattern that I think would be nice to follow, given many methods, would be something like:

```julia

"""
    cp_m(::A)
    cp_m(::B)
    cp_m(::C)

The isobaric specific heat capacity...
"""
function cp_m end

cp_m(...) = 
cp_m(...) = 
cp_m(...) = 
```

It's not a big deal, though. Maybe it'd be better to do that more systematically in a separate PR.

charleskawczynski:
> PhasePartition artificially clips values and I need a version that doesn't necessarily do that -- if it must be clipped I want to clip it in cloud microphysics, not in TD

Won't that lead to inconsistent phase partitions between the two packages? Also, why not just change the struct type to not perform clipping by default? We could then distribute the responsibility of clipped values, instead of making exceptions

oalcabes:
> We basically want to try get rid of all the clipping for moisture variables in `Thermodynamics.jl`, `CloudMicrophysics.jl` and `Atmos`. This is because all those additional clippings make it very difficult to write the derivatives for the implicit solver.

and so that we can implement a smarter way of limiting tendencies than just clipping them

PR #255, opened 2025-06-12
Bump patch and update news

oalcabes:


oalcabes (commented):
@JuliaRegistrator register

PR #256, opened 2025-07-01
gas_constant(q_tot, q_liq, q_ice)

trontrytel:
This PR provides an option to compute gas constant of moist air without using PhasePartiton

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -66,19 +66,25 @@ export q_vap_from_RH_liquid
 
 """
     gas_constant_air(param_set, [q::PhasePartition])
+    gas_constant_air(param_set, q_tot, q_liq, q_ice)
 
 The specific gas constant of moist air given
  - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
  - `q` [`PhasePartition`](@ref). Without this argument, the results are for dry air.
+ - `q_tot`, `q_liq` `q_ice` - specific liquid water contents for total water, liquid water and ice
 """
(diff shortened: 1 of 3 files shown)


src/relations.jl
```diff
@@ -66,19 +66,25 @@ export q_vap_from_RH_liquid
...
 The specific gas constant of moist air given
  - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
  - `q` [`PhasePartition`](@ref). Without this argument, the results are for dry air.
+ - `q_tot`, `q_liq` `q_ice` - specific liquid water contents for total water, liquid water and ice
 """
+@inline function gas_constant_air(param_set::APS, q_tot, q_liq, q_ice)
+    R_d = TP.R_d(param_set)
+    molmass_ratio = TP.molmass_ratio(param_set)
+    return R_d *
+           (1 + (molmass_ratio - 1) * q_tot - molmass_ratio * (q_liq + q_ice))
```
tapios, line 80: This is correct. However, it would be clearer to write it as follows, I think:
```suggestion
    R_v = TP.R_v(param_set)
    q_v = q_tot - q_liq - q_ice
    return R_d * (1 - q_t) * R_v * q_v
```

tapios (approved):
Thanks! LGTM. Consider the suggestion for a more transparent formulation.

src/relations.jl
trontrytel: Yes! I didn't want to change the formulation that was already coded in Thermodynamics. But this version is MUCH easier to read.

PR #258, opened 2025-07-15
Fix inverse molmass ratio bug

ychnli:
## Bug fixes
`partial_pressure_vapor` and `partial_pressure_dry` incorrectly had the molar mass ratio between dry air and water inverted. We use molmass_ratio $\equiv M_{air} / M_{H_2O} \approx 1.608$. 

## Affected functions
only the specific entropy functions were using these functions, and these were not previously used for diagnostics in ClimaAtmos so this bug went undetected. However, `partial_pressure_vapor` was being used in ClimaLand to compute vapor pressure deficit (VPD = e_sat - e) for the canopy model.

### DIFF
--- a/src/relations.jl
+++ b/src/relations.jl
@@ -3081,7 +3081,7 @@ The partial pressure of water vapor, given
 ) where {FT <: Real}
     molmass_ratio = TP.molmass_ratio(param_set)
     return p * (1 - q.tot) /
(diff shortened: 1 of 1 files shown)


tapios (approved):
LGTM. And please feel free to add a vapor_pressure_deficit function too (either in this PR or another one).

ychnli:
Thanks, @tapios. I'll add VPD in another PR

trontrytel (approved):
Thank you!

PR #259, opened 2025-07-16
Cleaned up/rewrote documentation, added functions needed for land, and cleaned up some logic

### LINKED ISSUE #177: `if`-statements may reduce performance on GPUs

### LINKED ISSUE #242: ClimaParams.jl required for docs example but missing from manifest
The code provided in the first page of the [docs](https://clima.github.io/Thermodynamics.jl/dev/#Simple-example) provides a good example of the minimum commands required to begin interfacing with Thermodynamics.jl. This example notes: ```using ClimaParams # needed in environment to load convenience parameter struct wrappers```
New users may be confused by this, however, as ClimaParams.jl is not included in the default Thermodynamics.jl manifest; running the example without ClimaParams.jl results in the following error:
```julia> TP.ThermodynamicsParameters(Float64)
ERROR: MethodError: no method matching Thermodynamics.Parameters.ThermodynamicsParameters(::Type{Float64})

Closest candidates are:
  Thermodynamics.Parameters.ThermodynamicsParameters(::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT, ::FT) where FT
   @ Thermodynamics ~/Thermodynamics.jl/src/Parameters.jl:29
  Thermodynamics.Parameters.ThermodynamicsParameters(; T_0, MSLP, p_ref_theta, cp_v, cp_l, cp_i, LH_v0, LH_s0, press_triple, T_triple, T_freeze, T_min, T_max, T_init_min, entropy_reference_temperature, entropy_dry_air, entropy_water_vapor, kappa_d, gas_constant, molmass_dryair, molmass_water, T_surf_ref, T_min_ref, grav, T_icenuc, pow_icenuc)
   @ Thermodynamics ~/Thermodynamics.jl/src/Parameters.jl:27

Stacktrace:
 [1] top-level scope
   @ REPL[1]:1
 [2] top-level scope
   @ none:1```

### PR DESCRIPTION
**Pull request by tapios**
This PR
* Renames `heavisided` to `ReLU` (because it's the ReLU function)
* Cleans up some docstrings in `relations.jl` to make them consistent with the rest
* Simplifies the logic for internal energy computation
* Adds methods for `vapor_pressure_deficit` 
* Adds tests for partial pressure and VPD functions 
* Made existing tests more physical and consistent. For example, fixed several issues with the use of `phase_type` that made some tests unphysical; changed temperature accuracies to absolute tolerances that are meaningful (many tests before had 10% relative tolerances, i.e., 20-30 K absolute tolerances, which are not meaningful).
* Renames `molmass_ratio` to `Rv_over_Rd`
* Renames `q_vap_saturation_from_density` to `q_vap_from_p_vap` (because it's not saturation specific); however, maintain backward compatibility
* Rewrites/updates the documentation extensively, to bring it in line with dycore paper and what we are actually doing (e.g., precip in working fluid). (Closes #242).
* Used `ifelse` where needed. Closes #177 

It should not change any behavior, except for adding new functionality (mostly needed for ClimaLand). The review can focus on relations.jl and the tests.

### CODE REVIEW

File: src/relations.jl [hunk 1]
```diff
@@ -1810,8 +1850,8 @@ Compute the temperature that is consistent with
  - `sat_adjust_method` the numerical method to use.
     See the [`Thermodynamics`](@ref) for options.
  - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
- - `h` specific_enthalpy
  - `p` air pressure
+ - `h` specific_enthalpy
```
szy21, line 1854:
```suggestion
 - `h` specific enthalpy
```

src/relations.jl
```diff
@@ -2554,13 +2602,19 @@ end
...
  - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
  - `T` temperature
  - `ρ` (moist-)air density
+
 and, optionally,
+
  - `q` [`PhasePartition`](@ref). Without this argument, the results are for dry air.
+
+The virtual potential temperature is defined as `θ_v = (R_m/R_d) * θ`, where `θ` is the
+potential temperature. It is the temperature a dry air parcel would need to have to
```
szy21, line 2616:
```suggestion
potential temperature. It is the potential temperature a dry air parcel would need to have to
```

src/relations.jl
```diff
@@ -2766,39 +2833,162 @@ given a phase partition, `q`.
...
+ - `T` air temperature
+ - `p` air pressure
+ - `q` [`PhasePartition`](@ref)
+"""
+@inline function vapor_pressure_deficit(
+    param_set::APS,
+    T::FT,
+    p::FT,
+    q::PhasePartition{FT},
+    Tᶠ = TP.T_freeze(param_set),
```
szy21, line 2952: Do we want users to be able to modify Tᶠ, or should we move Tᶠ inside the function?

NEWS.md
```diff
@@ -4,6 +4,21 @@ Thermodynamics.jl Release Notes
 main
 ----
 
+v1.xxx
```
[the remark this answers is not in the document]
tapios: 1.0 coming soon. A few more changes.

docs/make.jl
```diff
@@ -6,53 +6,52 @@ ENV["GKSwstype"] = "nul"
...
+#             execute = true,
+#         )
+#         elapsed = 1e-9 * (time_ns() - start_time)
+#         @info @sprintf("%s example took %s seconds to build.", example, elapsed)
+#     end
+# end
+# 
+# example_pages = [
+#     "Density from temperature, pressure, and humidity" => "literated/density_from_temperature_pressure_humidity.md",
+# ]
```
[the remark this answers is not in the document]
tapios: I'll remove it

src/relations.jl
tapios: Users do not necessarily need to modify Tᶠ. However, my understanding is that this pattern increases efficiency in repeated function calls?

szy21: Oh, I didn't know that. That's fine then.

test/relations.jl
```diff
@@ -287,152 +344,291 @@ end
         saturation_vapor_pressure(param_set, _T_triple, Ice()),
     ) ≈ 0.998 * q_tot / ρ_v_triple / ρ - 1
 
-    # energy functions and inverse (temperature)
+    # Helper function to test partial pressures
+    function test_partial_pressures(
+        T_test,
+        ρ,
+        q_partition,
+        test_name;
```
szy21, line 352: Is `test_name` used anywhere?

test/relations.jl
```diff
@@ -137,88 +184,97 @@ end
...
-    # specific latent heats
+    # Test latent heat calculations at reference temperature
     @test latent_heat_vapor(param_set, _T_0) ≈ _LH_v0
     @test latent_heat_fusion(param_set, _T_0) ≈ _LH_f0
     @test latent_heat_sublim(param_set, _T_0) ≈ _LH_s0
 
-    # saturation vapor pressure and specific humidity
+    # Test saturation vapor pressure and humidity calculations
     p = FT(1.e5)
     q_tot = FT(0.23)
```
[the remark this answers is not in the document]
tapios: I agree. I'll change it. (It's a holdover from older tests.)

test/relations.jl
tapios: Good point. I added a @testset that echos the name for help in finding errors.

test/relations.jl
```diff
@@ -287,152 +344,291 @@ end
...
+    q_pt = PhasePartition(FT(0.02), FT(0.002), FT(0.002))
     e_kin = FT(11)
     e_pot = FT(13)
-    @test air_temperature(param_set, _cv_d * (T - _T_0) - _R_d * _T_0) === FT(T)
-    @test air_temperature(
-        param_set,
-        _cv_d * (T - _T_0) - _R_d * _T_0,
-        PhasePartition(FT(0)),
-    ) === FT(T)
+    e_int_dry = internal_energy_dry(param_set, T)
```
szy21, line 492: Just a note. I think there was something useful in the previous test. The test would now fail to catch it if there is the same error in `internal_energy_dry` and `air_temperature`, for example, if they both use the wrong constant R_v * T_0 instead of R_d * T_0. I'm fine with this change though, as the better way to deal with this is to add a test for  the correctness of `internal_energy_dry` I think, which we don't have.

test/relations.jl
```diff
@@ -287,152 +344,291 @@ end
...
+    @test total_energy(param_set, FT(e_kin), FT(e_pot), FT(T)) ===
+          FT(e_kin) + FT(e_pot) + FT(e_int_dry)
+    @test total_energy(param_set, FT(e_kin), FT(e_pot), FT(T), q_pt) ≈
+          FT(e_kin) + FT(e_pot) + FT(e_int_moist)
+    @test total_energy(param_set, FT(0), FT(0), FT(T), q_pt) ≈ FT(e_int_moist)
+
+    # Test phase partitioning above freezing temperature
+    T_warm = FT(_T_freeze + 20)
     ρ = FT(1.0)
     q_tot = FT(0.21)
```
szy21, line 511: Use a better value?

test/relations.jl
tapios: Right, I just didn't like that we were manually recoding (i.e., copying and pasting) the definition from the source. I'd like the tests to be invariant under changes in the arbitrary constants, so what we should test is, e.g., consistency of energy and enthalpy. I'll add something.

test/relations.jl
```diff
@@ -476,12 +718,16 @@ end
...
+    q_pt = PhasePartition(q_tot)
+    @test TD.exner_given_pressure(param_set, p, q_pt) isa typeof(p)
+    @test TD.exner_given_pressure(param_set, p / 10, q_pt) ≈
+          TD.exner_given_pressure(param_set, p, q_pt) /
+          10^(gas_constant_air(param_set, q_pt) / cp_m(param_set, q_pt))
 
+    # Test humidity and mixing ratio calculations
     q_tot = 0.1
     q_liq = 0.05
     q_ice = 0.01
```
szy21, line 733: Use better values? (I don't care much about these though, as it is just testing mixing ratios.)

test/relations.jl
```diff
@@ -535,6 +780,11 @@ end
     for ArrayType in array_types
         FT = eltype(ArrayType)
         param_set = TP.ThermodynamicsParameters(FT)
+        _cp_d = TP.cp_d(param_set)
+        _LH_v0 = TP.LH_v0(param_set)
+        _eint_v0 = TP.e_int_v0(param_set)
+        _T_0 = TP.T_0(param_set)
+
```
szy21, line 787: Do we need anything other than _eint_v0?

test/relations.jl
```diff
@@ -893,24 +1143,34 @@ end
...
+                liquid_ice_pottemp.(param_set, ts_exact)
+            )
+        # For potential temperature, use a tolerance based on the iterative nature of the calculation
+        pot_temp_tolerances = 2.25 * atol_temperature  # Larger tol to allow for condensate effects
+        failing_pot_temp = findall(pot_temp_errors .> pot_temp_tolerances)
+        # if !isempty(failing_pot_temp)
+        #     @info "Potential temperature test failures" n=length(failing_pot_temp)
+        #     @show maximum(pot_temp_errors)
+        #     @show maximum(pot_temp_tolerances)
+        # end
```
szy21, line 1167: Do you want to remove these?

test/relations.jl
```diff
@@ -1032,6 +1292,10 @@ end
     for ArrayType in array_types
         FT = eltype(ArrayType)
         param_set = TP.ThermodynamicsParameters(FT)
+        _cp_d = TP.cp_d(param_set)
+        _LH_v0 = TP.LH_v0(param_set)
+        _eint_v0 = TP.e_int_v0(param_set)
+        _T_0 = TP.T_0(param_set)
```
szy21, line 1298: Do we need anything other than _eint_v0?

test/relations.jl
```diff
@@ -1127,48 +1399,45 @@ end
...
+            atol_energy_temperature .+ rtol_humidity * _eint_v0 .* q_tot
+        failing = findall(errors .> tolerances)
+        # if !isempty(failing)
+        #     @info "PhaseEquil_phq energy test failures" n=length(failing)
+        #     @show p[failing]
+        #     @show h[failing]
+        #     @show q_tot[failing]
+        #     @show errors[failing]
+        #     @show tolerances[failing]
+        # end
```
szy21, line 1420: Do you want to keep or remove this?

test/relations.jl
tapios: Need _cp_d too but the rest can go.

test/relations.jl
tapios: They were useful for debugging. Maybe I'll leave them for now.

test/relations.jl
tapios: No, good point. Removed the others.

PR #260, opened 2025-07-16
tag v0.12.15

juliasloan25:
Tag new version to use latest bugfix from #258

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@ Thermodynamics.jl Release Notes
 main
 ----
 
+v0.12.15
+--------
+- Fix inverse molmass bug. PR [258](https://github.com/CliMA/Thermodynamics.jl/pull/258)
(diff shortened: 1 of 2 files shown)


juliasloan25:
> Is the Land model ok to have it as patch release, even though it's behavior changing for them?

It seems like something may be wrong in some computation. I want to understand it before making the release. See the discussion on Slack [here](https://climate-machine.slack.com/archives/C05N77QDN7N/p1752708866700359)

kmdeck:
> > Is the Land model ok to have it as patch release, even though it's behavior changing for them?
> 
> It seems like something may be wrong in some computation. I want to understand it before making the release. See the discussion on Slack [here](https://climate-machine.slack.com/archives/C05N77QDN7N/p1752708866700359)

I fixed our land tests running with Thermodynamics main branch after speaking with @trontrytel. Regardless of how we fix it on our end, though, it is OK to merge this and make the release because the bug is on our end.

PR #265, opened 2025-07-25
Remove catches around freezing in saturation adjustment

tapios:
* Removes the catches we had around the freezing temperature to prevent saturation adjustment from oscillating when there is a discontinuity (Heaviside function) in phase partitioning at freezing. (Right now we have continuous phase partitioning and do not need these catches, which cause problems with automatic differentiation.)
* Larger changes to the test suite, which were necessary since it was testing for exact freezing temperatures, which will only be returned with the catch but not without it. I changed this to approximate (and more meaningful) tests. 
* Reduce code duplication in `saturation_adjustment.jl`
* Rename `condensate_shum` to `condensate_specific_humidity`
* A few smaller doc fixes


test/default_behavior_accuracy.jl
```diff
@@ -131,7 +136,7 @@ This file contains tests for saturation adjustment accuracy and convergence.
                     ρ,
                     _e_int,
                     q_tot,
-                    8,
+                    35,
```
szy21, line 139: Does this test now require more iterations to converge?

tapios:
Thanks. I don't have a buildkite account and cannot access this. Let me knowing there's anything for me to know. I know there are some type inference issues in SA, which I'll fix today.

src/saturation_adjustment.jl
```diff
@@ -93,59 +45,23 @@ See also [`saturation_adjustment`](@ref).
...
-    temperature_tol = _T_freeze * relative_temperature_tol
-    e_int_upper = e_int_sat(_T_freeze + temperature_tol / 2) # /2 => resulting interval is `temperature_tol` wide
-    if e_int < e_int_upper
-        e_int_lower = e_int_sat(_T_freeze - temperature_tol / 2) # /2 => resulting interval is `temperature_tol` wide
-        if e_int_lower < e_int # < e_int_upper
-            return _T_freeze
-        end
+    # Unsaturated Check
+    T_1 = max(_T_min, air_temperature(param_set, e_int, PhasePartition(q_tot)))
+    if T_1 >= _T_min && q_tot <= q_vap_saturation(param_set, T_1, ρ, phase_type)
```
szy21, line 50: Isn't T_1 >= _T_min always true as we do T_1 = max(_T_min, ...)?

tapios: Right, only the q_tot < q_sat condition really matters here. We can remove the temperature check.

szy21: There are some other places in the PR where we can remove the check too.

szy21:
What is the advantage of adding additional functions like `_saturation_adjustment_p_thermo_q` in `saturation_adjustment.jl`?

src/saturation_adjustment.jl
tapios: Will do

szy21 (approved):
Looks good, thanks! It would be great if you could squash the commits before merging.

PR #267, opened 2025-07-27
Remove dispatch on FT

szy21:
Removes dispatch on FT for dual number support that is required by automatic differentiation.

### DIFF
--- a/src/Thermodynamics.jl
+++ b/src/Thermodynamics.jl
@@ -67,7 +67,7 @@ include("printing.jl")
 @inline print_warning() = false
 
 # Default phase partition for dry air
-@inline q_pt_0(::Type{FT}) where {FT} = PhasePartition(FT(0), FT(0), FT(0))
+@inline q_pt_0(::APS{FT}) where {FT} = PhasePartition(FT(0), FT(0), FT(0))
 
 @inline solution_type() = RS.CompactSolution()
 include("DataCollection.jl")
--- a/src/air_dry_adiabatic.jl
+++ b/src/air_dry_adiabatic.jl
@@ -26,10 +26,10 @@ for an isentropic process: `p = p₀ * (1 - Φ/(θ * cₚ))^(cₚ/R)`, where `p�
 """
 @inline function air_pressure_given_θ(
     param_set::APS,
-    θ::FT,
-    Φ::FT,
+    θ,
+    Φ,
     ::DryAdiabaticProcess,
-) where {FT <: Real}
+)
     p0 = TP.p_ref_theta(param_set)
     _R_d = TP.R_d(param_set)
     _cp_d = TP.cp_d(param_set)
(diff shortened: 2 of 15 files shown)


szy21:
I am not sure if I changed everything correctly. I'll leave comments in the code for the questions I have.

szy21:
@dennisYatunin Could you take a look at `src/air_states.jl` and give me an example of what I should do for that file?

tapios:
We'll need to make the same changes in RootSolvers before we can test and merge this. I'm on it.

szy21:
@dennisYatunin This is not done yet (I still need to change air_states.jl and fix the tests), but please take a look and let me know if there is something I should change.

dennisYatunin:
I just took a more careful look. All of the changes in this PR are simply removing dispatch restrictions, except for this one:
```
@inline q_pt_0(::Type{FT}) where {FT} = PhasePartition(FT(0), FT(0), FT(0)) # original version
@inline q_pt_0(::APS{FT}) where {FT} = PhasePartition(FT(0), FT(0), FT(0)) # new version
```
Instead of passing `FT` to `q_pt_0`, you are now passing `param_set`. If for some reason the compiler were to ignore the `@inline` instruction (e.g., because it encountered a limit of some internal heuristic), more registers would be used to pass `param_set` instead of `FT`. I doubt this is actually happening, since the compiler generally follows `@inline` instructions, but just to be safe you can try going back to the original version.

If you go back to the original version, you can change every instance of `q_pt_0(param_set)` to `q_pt_0(eltype(param_set))`, so that you don't need to add annotations like `param_set::APS{FT} ... where {FT}`. More generally, you can use `FT = eltype(param_set)` in every function where you need `FT`, which would be a little more consistent that using annotations like `param_set::APS{FT}`, `p::FT`, `ρ::FT`, etc.

Also, when you test the effect of this on SYPD, make sure both versions are run on the same machine. Differences between processors alone could potentially lead to 3% slowdowns.

szy21:
> I just took a more careful look. All of the changes in this PR are simply removing dispatch restrictions, except for this one:
> 
> ```
> @inline q_pt_0(::Type{FT}) where {FT} = PhasePartition(FT(0), FT(0), FT(0)) # original version
> @inline q_pt_0(::APS{FT}) where {FT} = PhasePartition(FT(0), FT(0), FT(0)) # new version
> ```
> 
> Instead of passing `FT` to `q_pt_0`, you are now passing `param_set`. If for some reason the compiler were to ignore the `@inline` instruction (e.g., because it encountered a limit of some internal heuristic), more registers would be used to pass `param_set` instead of `FT`. I doubt this is actually happening, since the compiler generally follows `@inline` instructions, but just to be safe you can try going back to the original version.
> 
> If you go back to the original version, you can change every instance of `q_pt_0(param_set)` to `q_pt_0(eltype(param_set))`, so that you don't need to add annotations like `param_set::APS{FT} ... where {FT}`. More generally, you can use `FT = eltype(param_set)` in every function where you need `FT`, which would be a little more consistent that using annotations like `param_set::APS{FT}`, `p::FT`, `ρ::FT`, etc.
> 
> Also, when you test the effect of this on SYPD, make sure both versions are run on the same machine. Differences between processors alone could potentially lead to 3% slowdowns.

This is even worse, about 5% slower: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/475#0198c099-9002-4597-aa3b-0d99da1ef888. If you think we should make the change, I'm going to revert to the previous change and merge it. If you are concerned about the performance, could you ask someone (maybe Teja) to take a look? I would like to get this change in before we change the thermo state.

PR #270, opened 2025-07-29
add depot and set num precompile files

haakon-e:


nefrathenrici (approved):
Thank you! I didn't realize that the thermodynamics CI takes 45+ minutes to precompile.

PR #282, opened 2025-09-29
add specific enthalpy functions

szy21:


src/air_energies.jl
```diff
@@ -230,6 +230,23 @@ and total specific humidity, given
...
 
+"""
+    specific_dry_enthalpy_vapor(param_set, T)
+
+The specific dry enthalpy of vapor, given
+
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `T` temperature
+"""
+@inline function specific_dry_enthalpy_vapor(
```
szy21, line 241: @tapios Is this correct? I don't think I heard this definition before. Maybe I will define enthalpy and subtract Lv,0 in surface flux.

szy21:
@dennisYatunin I decided to not add FTs to the new functions - is that a good idea?

PR #283, opened 2025-10-07
set error_on_non_convergence to false

szy21:
I'm not sure if this will work though as in the code comments it says:

> Error on convergence must be the default behavior because this can result in printing very large logs resulting in CI to seemingly hang.

Maybe someone who had an SA error can try this branch?

### DIFF
--- a/src/Thermodynamics.jl
+++ b/src/Thermodynamics.jl
@@ -55,19 +55,16 @@ const APS = TP.AbstractThermodynamicsParameters
 # For printing literal strings on the gpu
 include("printing.jl")
 
-# Allow users to skip error on non-convergence
+# Allow users to print warning and throw errors on non-convergence
 # by importing:
 # ```julia
 # import Thermodynamics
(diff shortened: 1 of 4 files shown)


src/Thermodynamics.jl
```diff
@@ -55,19 +55,16 @@ const APS = TP.AbstractThermodynamicsParameters
 # For printing literal strings on the gpu
 include("printing.jl")
 
-# Allow users to skip error on non-convergence
+# Allow users to print waring and throw errors on non-convergence
```
tapios, line 58:
```suggestion
# Allow users to print warning and throw errors on non-convergence
```

tapios (approved):

haakon-e (commented):
I think this should be a major release.

szy21:
> I think this should be a major release.

It's just annoying to have breaking releases:) But yes I can change that.

szy21:
actually let me change another thing as we are going to have a breaking release

PR #293, opened 2026-01-04
Make Thermodynamics functional again and remove reliance on PhasePartition and ThermoStates 

### LINKED ISSUE #223: Remove dependency on KernelAbstractions
[ClimaTimeSteppers](https://github.com/CliMA/ClimaTimeSteppers.jl/issues/308) and Thermodynamics are the only two packages that bring in KernelAbstractions in the ClimaAtmos dependency tree. As far as I can tell, here, `KernelAbstractions` is only used with `KA.@print`, so it would be nice to remove this dependency.

### LINKED ISSUE #271: Simplify CI Environments
Thermodynamics.jl CI currently has 3 different Project.tomls used for different tests, requiring ~45 minutes of precompilation to use them all. We should simplify these into [one testing environment](https://pkgdocs.julialang.org/v1/creating-packages/#Test-specific-dependencies) using the "extras" from the package's Project.toml. I believe this would cut compilation time down to <15 minutes.

### LINKED ISSUE #273: Remove thermodynamic state, use more functional approach
# The Climate Modeling Alliance
### Purpose

We would like to reduce, and eventually remove, reliance on the atmospheric state constructor, which inflates memory use. The goal is to replace it by something that is functionally equivalent but simpler. The key insight is that we only need to compute the temperature from atmospheric model variables (e.g., internal energy, density). Everything else (e.g., partitioning of specific humidities) follows by simple computations. 

The temperature computation is different for `EquilMoistModel` (saturation adjustment) and `NonEquilMoistModel`/`DryModel` (direct computation). So we can dispatch over `MoistureModel` to compute temperature `T`, then use the so computed temperature to compute all specific humidities (which may be stored in a `humidities` tuple)

A design realizing this would look like this:
```
# Method types
"""
    MoistureModel

An abstract type for all moisture models.
"""
abstract type MoistureModel end

struct EquilMoistModel <: MoistureModel  end
struct NonEquilMoistModel <: MoistureModel  end
struct DryModel <: MoistureModel  end

"""
    air_temperature(::EquilMoistModel, param_set, ρ, e_int, q_tot; SAargs...)

Computes air temperature using saturation adjustment.

# Arguments
- `MoistureModel`: `EquilMoist()`
- `param_set`: Thermodynamics parameters
- `ρ`: density
- `e_int`: internal energy
- `q_tot`: total specific humidity
- `SAargs`: keyword arguments for saturation_adjustment
"""
function air_temperature(
    ::EquilMoist,
    param_set,
    ρ,
    e_int,
    q_tot;
    SAargs...,
)
    # Implementation using existing saturation_adjustment algorithms, allowing user control from ClimaAtmos over SA parameters such as the root solver method via SAargs
end

"""
    air_temperature(::NonEquilMoistModel, param_set, e_int, humidities)

Computes air temperature from internal energy and phase partition.

# Arguments
- `MoistureModel`: `NonEquilMoistModel()`
- `param_set`: Thermodynamics parameters
- `e_int`: internal energy
- `humidities`: A tuple containing q_tot, q_liq, and q_ice
"""
function air_temperature(
    ::NonEquilMoistModel,
    param_set,
    e_int,
    humidities,
)
    # Direct computation from internal energy and humidities
end

"""
    air_temperature(::DryModel, param_set, e_int)

Computes air temperature for dry air.

# Arguments
- `MoistureModel`: `DryModel()`
- `param_set`: Thermodynamics parameters
- `e_int`: internal energy
"""
function air_temperature(::DryModel, param_set, e_int)
    # Simplified computation for dry air
end
```
The `air_temperature` function can be extended with methods, dispatched over `InputVars`, that allow different input arguments, such as liquid ice potential temperature or moist static energy instead of internal energy

In addition, we'd need a function to compute the humidities tuple, which likewise dispatches over the `MoistureModel`:
```
"""
    humidities(::EquilMoistModel, param_set, T, ρ, q_tot)

Computes the specific humidities of the phases for equilibrium moist air.

# Arguments
- `MoistureModel`: `EquilMoistModel()`
- `param_set`: Thermodynamics parameters
- `T`: air temperature
- `ρ`: density
- `q_tot`: total specific humidity

# Returns
- `humidities = (q_tot, q_liq, q_ice)`
"""
function humidities(::EquilMoistModel, param_set, T, ρ, q_tot)
    # Compute humidities using liquid_fraction function and saturation excess
end
```
For `NonEquilMoistModel`, the humidities function is not needed; we can simply pass the named tuple `q = (; q_tot, q_liq, q_ice)` where needed. 

One may also consider a function `temperature_and_humidities` that computes the temperatures and humidities in one go. This could avoid one saturation vapor pressure computation.

### Benefits
* Reduced memory footprint: Lowers memory usage by eliminating the need to store a comprehensive thermodynamic state object for every grid point. Computations are performed on-the-fly from prognostic state variables.
* Improved composability and extensibility: The use of multiple dispatch on `MoistureModel` makes the system modular. Adding new physics models or supporting new input variables becomes trivial without altering existing code.
* Potential for increased performance: Julia's compiler may be able to optimize these dispatched function calls.

### Risks
* API transition: Managing the deprecation path for the old ThermodynamicState constructor requires care to ensure a smooth transition for downstream models.

### People and Personnel
- Lead: @tapios 
- Collaborators: @trontrytel 
- Reviewers: @szy21 

### Results and Deliverables

* Updated Therodynamics.jl package that is extended with the above functionality (retaining backward compatibility in a first pass)

### CC
@cmbengue @dennisYatunin

### LINKED ISSUE #292: Remove clipping specific humidities in phase partition

tapios:
This PR makes thermodynamics entirely functional (closes #273). The old `PhasePartition` and state methods are retained for the time being for backward compatibility, until ClimaAtmos, ClimaCoupler, ClimaLand etc. are updated. They are in files called `depr_*`, which we will delete then. (The code works without these files.)

There are a few (smaller) breaking changes:
- Renamed `specific_enthalpy*` to `enthalpy*`.
- Renamed `specific_entropy*` to `entropy*`.
- Renamed `dry_pottemp` to `potential_temperature`
- Removed `q_vap_saturation_generic`. Use `q_vap_saturation` instead.
- Removed `universal_gas_constant` (or `gas_constant`), `molar_mass_dryair` and `molar_mass_water` from thermo parameters.
- Removed `phase_type` from `relative_humidity`.
- Renamed `latent_heat_liq_ice` to `humidity_weighted_latent_heat`.

Also
- Simplifies CI (closes #271)
- Removed warnings/errors from saturation adjustment and replaced by converge flag (closes #223)
- Removes clippings (except in diagnostic quantities, such as relative humidity, closes #292)
- Fixes a few bugs (e.g., in the use of `liquid_fraction` in nonequilibrium conditions, derivatives used in saturation adjustment), which may affect model behavior.

### DIFF
--- a/test/correctness.jl
+++ b/test/correctness.jl
@@ -283,16 +283,17 @@ using the non-deprecated functional API (no `PhasePartition`/state types).
             # 1. Define checking logic
             function check_invariance(param_set_local, T_in, q_tot, q_liq, q_ice)
                 ρ = FT(1.2)
-                
+
                 # Internal Energy
                 e_int = TD.internal_energy(param_set_local, T_in, q_tot, q_liq, q_ice)
-                
+
                 # Recover Temperature
                 # air_temperature(param_set, phase, e_int, q_tot, ...)
                 # Using ρe() as phase indicator for internal energy
-                T_rec = TD.air_temperature(param_set_local, TD.ρe(), e_int, q_tot, q_liq, q_ice)
+                T_rec =
+                    TD.air_temperature(param_set_local, TD.ρe(), e_int, q_tot, q_liq, q_ice)
                 @test T_rec ≈ T_in
-                
+
                 # Enthalpy = e_int + R_m * T 
                 h = TD.enthalpy(param_set_local, T_in, q_tot, q_liq, q_ice)
                 R_m = TD.gas_constant_air(param_set_local, q_tot, q_liq, q_ice)
@@ -309,21 +310,24 @@ using the non-deprecated functional API (no `PhasePartition`/state types).
 
             # 3. Modify T_0 and run again
             # Construct a new parameter set with a different T_0
-            
+
             # Extract all fields as a NamedTuple
             nt = NamedTuple{fieldnames(TD.Parameters.ThermodynamicsParameters)}(
-                (getproperty(param_set, fn) for fn in fieldnames(TD.Parameters.ThermodynamicsParameters))
+                (
+                getproperty(param_set, fn) for
+                fn in fieldnames(TD.Parameters.ThermodynamicsParameters)
             )
-            
+            )
+
             # Create modified NamedTuple
             T_0_new = TP.T_0(param_set) + FT(10)
             nt_new = merge(nt, (; T_0 = T_0_new))
-            
+
             param_set_new = TD.Parameters.ThermodynamicsParameters{FT}(; nt_new...)
-            
+
             # Verify T_0 changed
             @test TP.T_0(param_set_new) ≈ T_0_new
-            
+
             # Run check again
             check_invariance(param_set_new, T_test, q_tot_test, q_liq_test, q_ice_test)
         end


src/ThermoTypes.jl
```diff
@@ -0,0 +1,85 @@
+export IndepVars
+export ρeq
+export peq
+export phq
+export pρq
+export pθ_liq_ice_q
```
szy21: do you prefer this over just `pθq`?

tapios: Yes, because I was concerned about confusion with dry θ (and did not like barely readable subscripts). I'm open to other suggestions though (especially now that I have new glasses ...)

szy21: I don't like subscripts either, so this is ok. Or maybe p_θ_liq_ice_q is better. I'm fine with pθliq too.

tapios: I simplified to `pe`, `ph`, `pθ_li` etc

src/air_dry_adiabatic.jl
```diff
@@ -1,8 +1,7 @@
 # Formulas assuming dry (unsaturated) adiabatic (i.e. isentropic) processes
 
 export DryAdiabaticProcess
-
-export air_pressure, air_temperature, air_pressure_given_θ
```
trontrytel: Is `air_pressure` and `air_temperature` meant to be internal only?

src/air_energies.jl
```diff
@@ -1,59 +1,69 @@
...
 export internal_energy_vapor
 export internal_energy_liquid
 export internal_energy_ice
 export internal_energy_sat
+export enthalpy
+export enthalpy_dry
+export enthalpy_vapor
+export enthalpy_liquid
+export enthalpy_ice
+export enthalpy_sat
```
trontrytel: I was wondering, because I always get confused by this a little. - Maybe we don't have to export `enthalpy_sat` and `internal_energy_sat`. I think they should only be used in saturation adjustment, because they internally assume phase partitioning. Or maybe even cleaner - we could only define them in those saturation adjustment places and not provide them  as standalone functions?

trontrytel:
This PR is so big GitHub only displays one file at a time to me :/

src/air_dry_adiabatic.jl
tapios: No, they are exported in the other files.

src/air_humidities.jl
```diff
@@ -1,244 +1,253 @@
...
+The formula is `q_vap = max(0, q_tot - q_liq - q_ice)`.
 """
-@inline ice_specific_humidity(q::PhasePartition) = q.ice
-
-"""
-    vapor_specific_humidity(q::PhasePartition)
-
-The vapor specific humidity, given a 
+@inline function vapor_specific_humidity(q_tot = 0, q_liq = 0, q_ice = 0)
+    return max(0, q_tot - q_liq - q_ice)
```
trontrytel: Do we want to keep clipping it?  Or maybe provide two functions `vapor_specific_humidity` and `vapor_specific_humidity_clipped` for testing?

src/air_humidities.jl
```diff
@@ -1,244 +1,253 @@
...
 
-The mixing ratio, given
- - `q` specific humidity
- - `q_tot` total specific humidity
+If the specific humidities are not given, the result is zero.
+The formula is `q_cond = q_liq + q_ice`.
 """
-@inline function shum_to_mixing_ratio(q, q_tot)
-    return q / (1 - q_tot)
+@inline function condensate_specific_humidity(q_liq = 0, q_ice = 0)
```
trontrytel: Is that function needed in many places? Maybe it's ok to just do the summation when needed

src/air_saturation_functions.jl
```diff
@@ -1,174 +1,360 @@
...
- - `param_set` an `AbstractParameterSet`
- - `T` temperature  
- - `LH_0` latent heat at reference temperature `T_0`
- - `Δcp` difference in isobaric specific heat capacity between the two phases
+# Reference
+Kaul et al. (2015), "Sensitivities in large-eddy simulations of mixed-phase Arctic stratocumulus
+clouds using a simple microphysics approach," *Monthly Weather Review*, **143**, 4393-4421,
+doi:[10.1175/MWR-D-14-00319.1](https://doi.org/10.1175/MWR-D-14-00319.1).
+"""
+@inline function liquid_fraction(param_set::APS, T, q_liq, q_ice)
```
trontrytel: It's a little bit counterintuitive to me to mix the prescribed temperature based definition with `q_liq / q_cond` one. I was blindly assuming that if `q_liq = 0` then I cannot have a non-zero cloud fraction. I understand that the second pathway is used in saturation adjustment, but maybe it would be better to use different names for them?

src/air_temperatures.jl
```diff
@@ -1,354 +1,322 @@
...
 
- - `q` [`PhasePartition`](@ref). 
- 
-When `q` is not provided, the results are for dry air.
+Note: if any of `q_tot`, `q_liq`, or `q_ice` are nonzero, the Exner function uses the
+moist exponent `R_m/cp_m` (reducing to the dry exponent `R_d/cp_d` in the dry limit).
 """
-@inline function air_temperature_given_ρθq(
+@inline function dry_pottemp_given_pressure(
     param_set::APS,
```
trontrytel: Same as above

src/air_temperatures.jl
```diff
@@ -1,354 +1,322 @@
...
-When `q` is not provided, the results are for dry air.
+If the specific humidities are not given, the result is for dry air.
+This method inverts [`enthalpy`](@ref) by solving for `T` given `h`.
 """
-@inline function air_temperature_from_enthalpy(
+@inline function air_temperature(
     param_set::APS,
-    h,
-    q::PhasePartition = q_pt_0(param_set),
+    ::ph,
```
szy21: same here

src/air_temperatures.jl
```diff
@@ -1,354 +1,322 @@
...
+If the specific humidities are not given, the result is for dry air.
 
- - `q` [`PhasePartition`](@ref). 
- 
-When `q` is not provided, the result is the dry-air potential temperature.
+Note: if any of `q_tot`, `q_liq`, or `q_ice` are nonzero, the Exner function uses the
+moist exponent `R_m/cp_m` (reducing to the dry exponent `R_d/cp_d` in the dry limit).
 """
-@inline function liquid_ice_pottemp(
+@inline function dry_pottemp(
```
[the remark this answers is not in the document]
szy21: I think we need q to calculate pressure and exner?

test/TestedProfiles.jl
```diff
@@ -0,0 +1,343 @@
...
+Notes:
+- These profiles are intended for **testing**; some are in thermodynamic equilibrium
+  (`EquilMoistProfiles`) while others are intentionally non-equilibrium (`NonEquilMoistProfiles`).
+- `TemperatureProfiles` return a temperature-like quantity which may be interpreted as
+  virtual temperature, depending on the profile implementation.
+"""
+module TestedProfiles
+
+using Thermodynamics
+const TD = Thermodynamics
```
szy21:
```suggestion
import Thermodynamics as TD
```
(and same for other const)

test/correctness.jl
```diff
@@ -1,845 +1,151 @@
...
-        q_pt = PhasePartition(FT(0.02), FT(0.005), FT(0.001))
-        h_tot_moist = total_specific_enthalpy(param_set, e_tot, T, q_pt)
-        R_m_moist = gas_constant_air(param_set, q_pt)
-        @test h_tot_moist ≈ e_tot + R_m_moist * T
+            # Saturated: vapor is at saturation and condensate is nonnegative
+            q_tot = FT(1.3) * q_sat
+            (q_liq, q_ice) = TD.condensate_partition(param_set, T, ρ, q_tot)
+            @test q_liq ≥ 0
+            @test q_ice ≥ 0
+            @test q_liq + q_ice ≤ q_tot + FT(100) * eps(FT)
```
szy21:
```suggestion
            @test q_liq + q_ice ≤ q_tot
```

src/air_humidities.jl
tapios: I removed the clippings for now.

src/air_saturation_functions.jl
```diff
@@ -178,109 +364,171 @@ The saturation specific humidity, given
...
+    p,
+    q_tot = 0,
+    q_liq = 0,
+    q_ice = 0,
 )
-    q_vap_sat = q_vap_from_p_vap(param_set, T, ρ, p_vap_sat)
-    return max(0, q.tot - q_vap_sat)
+    Tᶠ = TP.T_freeze(param_set)
+    above_freezing = T > Tᶠ
+    es = ifelse(
```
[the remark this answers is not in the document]
tapios: Added additional VPD methods that take a Phase (Liquid() or Ice()) argument.

src/air_temperatures.jl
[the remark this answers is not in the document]
tapios: I would have liked to rename dry_pottemp to pottemp (or potential_temperature), since the exponents are the moist exponents. It's just used in many places outside TD. What's your take: should we rename it to avoid this confusion?

tapios:
> I went through all the files in source that are not labelled as depreciated. Tomorrow I'll update CM.jl to the new version of Thermodynamics and see if any tests are flagged.

Thanks so much! Really helpful comments.

src/air_temperatures.jl
tapios: Fair point, as above. But I prefer the clean dispatch over independent variable types.

trontrytel (approved):
I tested the new version with CloudMicrophysics.jl and I see no behavior changes. Thank you!

PR #296, opened 2026-01-10
Backward compatibility wrappers

tapios:
Adds wrappers with old function names for backward compatibility of ClimaAtmos, ClimaLand etc. with this version of Thermodynamics.jl

This PR may allow us to release this version of TD and then make the downstream changes, before removing the deprecated code.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -152,6 +152,24 @@ Thermodynamics.DataCollection
 
 These functions are deprecated and will be removed in a future release.
 
+### Backward Compatibility Wrappers
+
+These wrappers exist for backward compatibility with older versions of the package.
+
+```@docs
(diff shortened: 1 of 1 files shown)


szy21 (approved):
Thanks. I'm fine with not maintaining the backward compatibility though. I plan to update ClimaAtmos with the new Thermodynamics as soon as SurfaceFluxes is done.

tapios:
> Thanks. I'm fine with not maintaining the backward compatibility though. I plan to update ClimaAtmos with the new Thermodynamics as soon as SurfaceFluxes is done.

I thought this may allow us to work more piecemeal in the interim, working in atmos, coupler, land in sequence. I do want to delete these wrappers and other redundant code as soon as we can though.

PR #348, opened 2026-08-10
Thermo updates: SA convergence improvements, expanded tests, docs cleanup

tapios:
Several updates:
- Improved convergence of saturation adjustment with more accurate derivatives in Newton solver
- GPU performance improvements by reducing exp/log function calls (hoisting them out of SA loop)
- Expanded test coverage
- Corrected several inconsistencies in docs

Bumps minor release (SA convergence changes may affect ClimaAtmos behavior slightly).


src/saturation_adjustment.jl
```diff
@@ -825,16 +474,20 @@ function saturation_adjustment_fixed_iters end
...
+    saturated = _is_saturated(param_set, T_0, ρ, q_tot)
+
+    T = T_0
+    ΔT = zero(T)
+    for _ in 1:maxiter
+        e_val = internal_energy_sat(param_set, T, ρ, q_tot, Val(false))
+        de_int_dT = ∂e_int_∂T_sat_ρ(param_set, T, ρ, q_tot, Val(false))
+        (T, ΔT) = _newton_update(param_set, T, (e_int - e_val) / de_int_dT)
     end
+    (T, ΔT) = _select_solution(saturated, T, ΔT, T_0)
```
szy21: Here and in other places, it returns T_init_min instead of T_unsat for unsaturated conditions with T < T_init_min. Is this what we want?

src/saturation_adjustment.jl
```diff
@@ -440,380 +226,243 @@ function saturation_adjustment(
...
-- With `maxiter = 2`, temperature accuracy is better than 0.1 K for typical atmospheric
-  conditions (T < 320 K).
-- This is an internal helper function. For the public API, use [`saturation_adjustment`](@ref)
-  with `forced_fixed_iters=true` as a positional argument.
+This is a test on the iteration itself, not on a residual: the fixed-iteration solvers take
+a set number of steps and never evaluate a stopping criterion. A small final increment means
+the iteration has settled; a large one means `maxiter` ran out before it did.
 """
-function saturation_adjustment_fixed_iters end
+@inline function _fixed_iters_converged(T, ΔT)
```
szy21, line 478: From Claude, I didn't check if this is true but can check if needed.

> The last-increment convergence test lags one iteration: a saturated 300 K state solved with maxiter = 5 has a temperature error of 9.3e-10 K, yet reports converged = false because the final Newton increment (~1e-4 K) still exceeds the tolerance sqrt(eps(Float64))·T ≈ 4.5e-6 K. With the default maxiter = 2, essentially every saturated state reports failure.

src/saturation_adjustment.jl
```diff
@@ -440,380 +226,243 @@ function saturation_adjustment(
...
-    sa_result = saturation_adjustment_fixed_iters(param_set, ph(), p, h, q_tot, maxiter)
-    return (; sa_result.T, sa_result.q_liq, sa_result.q_ice)
+@inline function _newton_update(param_set::APS, T, ΔT_raw)
+    FT = eltype(param_set)
+    T_init_min = TP.T_init_min(param_set)
+    # Wide enough never to bind near a solution, small enough to keep a diverging
+    # iterate within the range where the saturation functions remain informative.
+    ΔT_max = FT(50)
+    ΔT = clamp(ΔT_raw, -ΔT_max, ΔT_max)
+    T_new = max(T_init_min, T + ΔT)
```
szy21, line 397: From Claude:

> When Newton steps repeatedly drive T below T_init_min, _newton_update clamps to 150 K and returns ΔT = T_new - T = 0, so _fixed_iters_converged sees a zero increment and reports success for a state the solver never actually solved. 

I'm not sure if it matters, as that's probably a very unphysical state.

src/saturation_adjustment.jl
```diff
@@ -1524,24 +1389,20 @@ Uses `∂q_vap_sat/∂T|_p = q_vap_sat * L / (R_v T²)` (Clausius-Clapeyron at c
...
-    e_liq = internal_energy_liquid(param_set, T)
-    e_ice = internal_energy_ice(param_set, T)
+@inline function ∂e_int_∂T_sat_p(
+    param_set::APS,
+    T,
+    p,
+    q_tot,
+    unsat_branch::Val = Val(true),
+)
+    vars = _saturation_derivative_vars_p(param_set, T, p, q_tot)
```
szy21: Also from Claude:

> When the residual is evaluated with Val(false) (unclamped saturation excess, i.e. negative q_c allowed in the icenuc ramp), the derivative helpers ∂e_int_∂T_sat_ρ / ∂T_sat_p never forward that flag into _saturation_derivative_vars(_p) — they always compute with clamped q_c = 0. Measured against finite differences, this gives a 1.0–1.6% derivative error for iterates in the subsaturated ramp region (e.g. T = 270–273 K, q_tot = 0.003, ρ = 0.9). That degrades the quadratic convergence the maxiter = 2 accuracy claims rest on, and contradicts the documented "residual and derivative are a consistent continuation" design. 

It seems to make sense to me. Not sure if it matters. If it's the case the fix it suggests is pretty simple:

> 1. Give the fixed-ρ helper a clamped parameter (the fixed-p helper already has one). At [line 1263](vscode-webview://1kbod9ei73tv5uti4p9m3ke2m2ak538e7q9istmoq8jsdafn9ktg/src/saturation_adjustment.jl#L1263):

```
@inline function _saturation_derivative_vars(
    param_set::APS,
    T,
    ρ,
    q_tot,
    q_vap_sat,
    ::Val{:ρ},
    clamped::Val = Val(true),   # new
)
    λ = liquid_fraction_ramp(param_set, T)
    q_c = saturation_excess(param_set, T, ρ, q_tot, clamped)   # was: no flag
    ...
```

> 2. Forward unsat_branch at the six call sites.

Function | Line | Change
-- | -- | --
∂e_int_∂T_sat_ρ | 1330 | _saturation_derivative_vars(param_set, T, ρ, q_tot, q_vap_sat, Val(:ρ), unsat_branch)
∂e_int_∂T_sat_p | 1399 | _saturation_derivative_vars_p(param_set, T, p, q_tot, unsat_branch)
∂h_∂T_sat_p | 1423 | same as above
∂θ_li_∂T_sat_p | 1447 | same as above
∂θ_li_∂T_sat_ρ | 1545 | same as line 1330
∂p_∂T_sat_ρ | 1599 | same as line 1330

src/saturation_adjustment.jl
szy21: Do we want to clip T with ReLU(T) as in other functions?

szy21: same here, do we want to clip T?

src/saturation_adjustment.jl
```diff
@@ -1298,12 +984,155 @@ end
...
+
+    # Logarithmic derivative of the mixed-phase saturation vapor pressure
+    L = latent_heat_mixed(param_set, T, λ)
+    ∂lnp_∂λ = log_saturation_vapor_pressure_ratio(param_set, T)
+    ∂lnp_v_sat_∂T = L / (R_v * T^2) + ∂lnp_∂λ * ∂λ_∂T
+
+    # Guard the denominator the same way q_vap_saturation_from_pressure_calc does, so that
+    # p approaching p_v_sat cannot produce a division by zero inside a kernel.
+    Δp = p - p_v_sat
+    amplification = ifelse(Δp ≥ ϵ_numerics(FT), p / Δp, one(Δp))
```
szy21: should it be 0 when Δp < ϵ as q_vap_sat is always 1?

src/saturation_adjustment.jl
tapios: Ah, good point. No, it's not what we want. Fixed it.

szy21 (approved):

src/saturation_adjustment.jl
tapios: True. Replaced the eps-derived tolerance with a fixed 1e-5·T (~3 mK), which means the same thing in both precisions. This now agrees in tests with the `converged` status that is reported.

src/saturation_adjustment.jl
tapios: True. _newton_update now returns the increment Newton requested, not the clamped output. Also removed the T_init_min floor that led to this.

src/saturation_adjustment.jl
tapios: Implemented as suggested and verified derivatives against finite differences.

src/saturation_adjustment.jl
tapios: Yes, fixed.

src/saturation_adjustment.jl
tapios: Fixed consistently.

tapios: Ditto

